PR #5, opened 2019-05-02
Fix cube frontFaces

Kangz:


austinEng:
Thanks, Victor fixed this in #4.

PR #14, opened 2019-09-27
Two new cube samples

shrekshao:
Some trivial attempts I made during learning WebGPU

Screen Shot 2019-09-27 at 2 24 41 PM
Screen Shot 2019-09-27 at 2 25 01 PM

### DIFF
--- a/instanced_cubes.html
+++ b/instanced_cubes.html
@@ -1,0 +1,318 @@
+<!DOCTYPE html>
+<html>
+
+<head>
+    <meta name="viewport" content="width=600">
+    <title>WebGPU Instanced Cubes demo</title>
+    <script src="third_party/gl-matrix-min.js"></script>
+</head>
+
+<body>
+    <div id="not-supported" style="display: none">WebGPU not supported! Please visit <a href="//webgpu.io">webgpu.io</a> to see the current implementation status.</div>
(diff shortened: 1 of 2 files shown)


austinEng:
Awesome! Thanks Shrek. The instancing one is a great example! Is the two_cubes sample significantly different from other simple cube example? It looks roughly the same except we have two model matrices and draw it twice. I think we only need to add the instanced example.

shrekshao:
I think the two_cubes sample mainly demonstrate scenario using more than one uniformBindGroups (instanced only have one). 

The offset has to be 256byte aligned https://github.com/austinEng/webgpu-samples/pull/14/files#diff-8edc081d107d2dce08f854f495cb164bR240

I consider it helpful as in GL I usually have 1 uniform matrix and call uniform4fv to reset the same uniform between two drawcall, while here you cannot setSubData for the same variable between recording of drawcall.

austinEng:
Ah, that makes sense that it shows a significant difference between uploading data in WebGPU and WebGL. Thanks!

shrekshao:
Maybe I should archive this https://github.com/austinEng/webgpu-samples/pull/14#issuecomment-537287005 and add this description to readme in the future. Thanks for reviewing.

Issue #15, opened 2019-10-05
Fix coordinate systems to match spec


kainino0x:
Now that we've decided on the coordinate systems in the spec, we need to update these samples. https://github.com/gpuweb/gpuweb/issues/458

austinEng:
Fixed in #19

PR #16, opened 2019-10-07
Fix coordinate

shrekshao:
#15


fractal_cube.html
```diff
@@ -59,46 +59,46 @@
         const cubeVerticesArray = new Float32Array([
             // float4 position, float4 color, float2 uv
             1, -1, 1, 1, 1, 0, 1, 1, 1, 1,
-            -1, -1, 1, 1, 0, 0, 1, 1, 0, 1,
             -1, -1, -1, 1, 0, 0, 0, 1, 0, 0,
+            -1, -1, 1, 1, 0, 0, 1, 1, 0, 1,
```
austinEng: Changing the vertex positions does work, but I don't think it's the best way to fix this. What changed is the 2D coordinate systems so we need to fix how we're projecting these vertices into 2D space.

It's probably best done by removing the negative sign on `-aspect` in
`mat4.perspective(projectionMatrix, (2 * Math.PI) / 5, -aspect, 1, 100.0);` below

shrekshao: Make sense. So you workaround this before by setting the aspect to negative.

PR #18, opened 2019-10-09
Add webpack build to compile Typescript sources and generate nicer UI

austinEng:
Existing .html files will be removed once an automated build is running.


austinEng:
@kainino0x @shrekshao Just added you as collaborators. This is a huge mega-patch so we can migrate to use Typescript, generate stuff for a nicer UI, factor helper functions, etc. A ton of it is just copy-paste samples into `.js` files.

shrekshao:
Just want to make sure if there's also generated javascript included. If I were a dev learning webgpu  it's likely I want to copy 'n paste javascript code.

austinEng:
Hm, try out the build and let me know what you think.. it almost works right out of the box?

There's a couple of helpers in places but maybe instead we can get the build to inline source maps into each sample.

austinEng:
I have a thing in there that fetches the source code and dumps it on the page. I think we could eventually make that do a transform to vanilla JS

kainino0x:
The compiled samples are going to be JS anyway. So if you view source on github.io, you'll see JS (probably with TS source maps). If you look at github.com, you'll see TS. Maybe just add a note that to see the generated js, look in the gh-pages branch or github.io?

PR #19, opened 2019-10-10
Fix y-flip due to coordinate system change in WebGPU #15

austinEng:


austinEng:
fyi @shrekshao

shrekshao:
Hmm...... I don't think the current matrix setup is correct... (though I haven't figured out what should be the correct one yet, almost forgot everything about math)

WebGPU is left-handed. gl-matrix is written against WebGL so it's righthanded. Matrices generated by the lib cannot be used directly with WebGPU.

Did coordinate system discussion in WebGPU meeting talk about reusability of existing javascript graphics lib?

Screen Shot 2019-10-10 at 1 06 42 PM

Screen Shot 2019-10-10 at 1 07 42 PM

shrekshao:
@austinEng @kainino0x 
Link something (didn't solve problem) https://stackoverflow.com/questions/1263072/changing-a-matrix-from-right-handed-to-left-handed-coordinate-system
to start some discuss in case either of you know the correct solution in heart.

shrekshao:
I think the solution for now without changing the model (that's likely be in right-handed space and used to work with GL) is
`mat4.scale(projectionMatrix, projectionMatrix, vec3.fromValues(1, -1, 1));`
and changing the front face to be `cw`

kainino0x:
Depending on how you're generating the matrix with gl-matrix (I haven't
looked at the code recently), it's possible you can get the right result by
negating some parameter, such as the fovy or aspect ratio.

However the scale by 1,-1,1 definitely sounds correct to me so no problem
there.

PR #21, opened 2019-10-10
Fix texture y-flip; Fix cube uv typo.

shrekshao:
I believe the texture drawn on canvas need to flip to make sure `texture(t, uv)` provides the correct result.

Fix a uv vertices value typo by the way.

(Is gh-page manually build?)


src/helpers.ts
```diff
@@ -20,6 +20,8 @@ export async function createTextureFromImage(device: GPUDevice, src: string, usa
   imageCanvas.height = img.height;
 
   const imageCanvasContext = imageCanvas.getContext('2d');
+  imageCanvasContext.translate(0, img.height);
+  imageCanvasContext.scale(1, -1);
```
austinEng: I'm concerned about this because the flip of the canvas could cause additional copies when the canvas is composited. @kainino0x might know more about this.

That being said, I think we should probably update the samples to use the API "correctly" instead of using it like it's WebGL and fixing the flip after.

austinEng: Sorry. I just realized this is the image upload canvas and not the drawing canvas. LGTM

kainino0x: btw, .scale() is a state setting function on canvas 2d. So what it does is tells the future drawImage to use that state.

We should get a TODO in here to use copyImageBitmapToTexture when it's available.

PR #23, opened 2019-10-11
Update setVertexBuffers -> setVertexBuffer

austinEng:
Note: don't merge until Canary updates

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
     "gl-matrix": "^3.1.0"
   },
   "devDependencies": {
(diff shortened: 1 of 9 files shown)


austinEng:
Awesome. Updated the types. Also I realized that this is already updated in Canary, so merging now.
Things will still be broken until https://chromium-review.googlesource.com/c/chromium/src/+/1856618 is in.

shrekshao:
lgtm. Thanks for the update. (Don't know how to review...)

PR #24, opened 2019-11-06
Add live shader editing

austinEng:


src/examples/computeBoids.ts
```diff
@@ -141,12 +141,20 @@ export async function init(canvas: HTMLCanvasElement) {
...
       entryPoint: "main"
     },
     fragmentStage: {
       module: device.createShaderModule({
         code: glslang.compileGLSL(fragmentShaderGLSL, "fragment"),
+
+        // @ts-ignore
+        source: fragmentShaderGLSL,
+        transform: source => glslang.compileGLSL(source, "fragment"),
       }),
```
kainino0x: probably can use `{ ... } as GPUShaderModuleDescriptor` instead of @ts-ignore.

PR #30, opened 2020-01-02
Fixed row pitch padding

### LINKED ISSUE #29: Broken row pitch computation or implementation
I was trying to understand what the row pitch computation was doing so I built the textured cube sample locally and tried with a different image. I simply scaled down the provided image from 512 x 512 px to 500 x 500 px.  Here is the result:

Untitled

I'm not sure what is going on as the row pitch parameter is not described in the WebGPU
Editor’s Draft. Is this a normal behavior in the current state of implementation?

frguthmann:
The previous code was not loading images correctly when padding was required to fulfill row pitch requirements.

Thanks @vorg for finding the problem.

Fixes #29

### DIFF
--- a/src/helpers.ts
+++ b/src/helpers.ts
@@ -32,13 +32,15 @@ export async function createTextureFromImage(device: GPUDevice, src: string, usa
     data = imageData.data;
   } else {
     data = new Uint8Array(rowPitch * img.height);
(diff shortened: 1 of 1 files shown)


frguthmann:
No problem, thank you for your work on WebGPU :).

PR #39, opened 2020-04-09
Update WebGPU API usage

### LINKED ISSUE #38: Update to the lastest API
There was a number of adjustments in the API, like `bindings` -> `entries` change, that would be nice to see here.

kainino0x:
Tested both before and after this patch against my unlanded Chromium changes to add the new API surface with non-breaking deprecations.

Fixes #38.

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
   },
   "devDependencies": {
     "@types/codemirror": "^0.0.88",
-    "@webgpu/types": "^0.0.22",
+    "@webgpu/types": "^0.0.23",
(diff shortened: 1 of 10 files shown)


austinEng:
Not sure what's up with Travis. The job finished very quickly but never reported the result.

kainino0x:
Had that same issue on CTS last week. Probably travis infra issue that's fixed by now.

kainino0x:
The travis issue seems to have prevented this change from deploying to github.io :\

PR #50, opened 2020-08-24
Add a WGSL version of HelloTriangle

Kangz:
@austinEng WDYT? Should we have both versions in parallel like this or separate files for SPIRV and WGSL examples. Also what about the helper functions, should they be moved to a common file?


Kangz:
Travis doesn't seem to trigger for this PR :/

kainino0x:
It built but just didn't report back to GitHub: https://travis-ci.org/github/austinEng/webgpu-samples/builds/723019730
Been having problems like this with Travis consistently for many months :\

PR #58, opened 2020-09-11
Updating texturedCube example to use WGSL shaders

ponitka:
Works with Tint with this change: https://dawn-review.googlesource.com/c/tint/+/28125

### DIFF
--- a/src/examples/texturedCube.ts
+++ b/src/examples/texturedCube.ts
@@ -6,7 +6,7 @@ import glslangModule from '../glslang';
 export const title = 'Textured Cube';
 export const description = 'This example shows how to bind and sample textures.';
 
-export async function init(canvas: HTMLCanvasElement) {
+export async function init(canvas: HTMLCanvasElement, useWGSL: boolean) {
   const adapter = await navigator.gpu.requestAdapter();
   const device = await adapter.requestDevice();
   const glslang = await glslangModule();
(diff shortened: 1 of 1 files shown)


src/examples/texturedCube.ts
```diff
@@ -203,3 +203,41 @@ void main() {
...
+[[location 0]] var<out> fragUV : vec2<f32>;
+[[location 1]] var<out> fragPosition: vec4<f32>;
+
+fn vtx_main() -> void {
+  fragPosition = 0.5 * (position + vec4(1.0));
+  Position = uniforms.modelViewProjectionMatrix * position;
+  fragUV = uv;
+  return;
+}
+entry_point vertex as "main" = vtx_main;
```
kainino0x, line 227: nit: if `vtx_main` were named `main` this could just be `entry_Point vertex = main;`

kainino0x (approved):
LGTM but I guess we can wait on the spec before landing.

src/examples/texturedCube.ts
```diff
@@ -203,3 +203,41 @@ void main() {
...
+`,
+  fragment: `
+[[binding = 1, set = 0]] var<uniform> mySampler: sampler;
+[[binding = 2, set = 0]] var<uniform> myTexture: texture_sampled_2d<f32>;
```
kainino0x, line 231: Just realized these have `=` while the above ones don't. Pretty sure there should be no `=`. (Syntax is probably going to change soon anyway though.)

ponitka:
> Works with Tint with this change: https://dawn-review.googlesource.com/c/tint/+/28125

Whoops, that was because the sample was still running GLSL shaders.

austinEng:
Thanks for the contribution! :) I ended up cherry-picking your commits and merging them in #60 with a bunch of other WGSL changes.

PR #68, opened 2020-11-29
fix: Update source link to use master -> main branch

lukehesluke:
As the default branch has changed from master -> main.


austinEng (approved):
Thanks!

PR #74, opened 2021-01-13
s/texture_sampled_2d/texture_2d

ben-clayton:
These texture types were renamed in October 2020:
https://github.com/gpuweb/gpuweb/pull/1180


austinEng (approved):
Thanks!

PR #80, opened 2021-02-24
replace deprecated apis and fix several warnings

HypnosNova:
All demos are tested successful in newest chrome canary.

Fix following warnings: 
image

### DIFF
--- a/src/pages/samples/videoUploading.ts
+++ b/src/pages/samples/videoUploading.ts
@@ -95,7 +95,7 @@ async function init(canvas: HTMLCanvasElement) {
     size: {
       width: video.videoWidth,
       height: video.videoHeight,
-      depth: 1,
+      depthOrArrayLayers: 1,
     },
     format: 'rgba8unorm',
     usage: GPUTextureUsage.COPY_DST | GPUTextureUsage.SAMPLED,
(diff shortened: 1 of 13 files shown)


src/pages/samples/videoUploading.ts
```diff
@@ -117,10 +117,10 @@ async function init(canvas: HTMLCanvasElement) {
 
   return function frame() {
     createImageBitmap(video).then((videoFrame) => {
-      device.defaultQueue.copyImageBitmapToTexture(
+      device.queue.copyImageBitmapToTexture(
         { imageBitmap: videoFrame, origin: { x: 0, y: 0 } },
         { texture: videoTexture },
-        { width: video.videoWidth, height: video.videoHeight, depth: 1 }
+        { width: video.videoWidth, height: video.videoHeight, depthOrArrayLayers: 1 }
```
kainino0x, line 123: optional nit: `depthOrArrayLayers` defaults to 1 so we could omit these. Especially nice for this one because it prevents the dict from breaking over multiple lines.

Could also be `[video.videoWidth, video.videoHought]` but no need to make that change.

kainino0x (approved):
nice, thank you!

There are a few formatting issues, could you get eslint to autofix them? thanks :)

src/pages/samples/videoUploading.ts
HypnosNova: I found that in newest @webgpu/types,  "depthOrArrayLayers" is not an optional property. 
I will update it in @webgpu/types later

kainino0x: Oops, I totally didn't notice this (until I happened to see it earlier today). Sorry about that!

It's fine to leave these in, in this PR.

kainino0x: (or wait a day and I'll have landed fixes in the types and published a new version to npm)

austinEng (approved):

src/pages/samples/videoUploading.ts
austinEng: +1, thanks for the fix! feel free to land it now with the current types, or wait for the update.

HypnosNova:
image
Since ts build is in problem because of the type checking, we'd better publish the newest @webgpu/types first and then I update the package.json

kainino0x:
Types 0.0.41 is now published, it should work. fyi in case it comes up: you might need `Required<GPUExtent3DDict>` https://github.com/gpuweb/types/pull/55#issuecomment-786265148

PR #83, opened 2021-03-22
Update following the latest API changes

austinEng:
Updates vertex formats, usage of GPUBindGroupLayoutEntry, and GPURenderPipelineDescriptor. Adds `u` suffix to integer comparisons in compute_boids.


austinEng:
@toji fyi

PR #85, opened 2021-03-31
Reversed depth buffer sample

shrekshao:


src/pages/samples/reversedZ.ts
```diff
@@ -0,0 +1,973 @@
...
+enum DepthBufferMode {
+  Default = 0,
+  Reversed,
+
+  COUNT,
+}
+
+const depthCompareFuncs: GPUCompareFunction[] = ['less', 'greater'];
+const depthLoadValues = [1.0, 0.0];
+
```
austinEng, line 66: may be more readable if we had a:

```
const depthBufferModes: DepthBufferMode[] = [DepthBufferMode.Default, DepthBufferMode.Reversed]
const depthCompareFuncs = {
  [DepthBufferMode.Default]: 'less',
  [DepthBufferMode.Reversed]: 'greater',
};
```

Then, we don't need COUNT, and the for-loops in the frame loop could just be `for (const depthBufferMode of depthBufferModes) {`

austinEng (approved):
Overall, really nice demo that demonstrates the precision difference. I think it would be good to have more comments throughout to explain what each pipeline/texture is used for, etc. Also, to call out which parts of the code are for your debug views, and which parts are for the actual reverseZ method.

src/pages/samples/reversedZ.ts
```diff
@@ -0,0 +1,973 @@
...
+const yCount = 5;
+const numInstances = xCount * yCount;
+const matrixFloatCount = 16; // 4x4 matrix
+const matrixStride = 4 * matrixFloatCount;
+
+const depthRangeRemapMatrix = mat4.create();
+depthRangeRemapMatrix[10] = -1;
+depthRangeRemapMatrix[14] = 1;
+
+// https://github.com/toji/gl-matrix/commit/e906eb7bb02822a81b1d197c6b5b33563c0403c0
```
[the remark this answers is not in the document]
shrekshao: Last release is April 1 2020 so this update is not in distribution now (maybe I can just ask @toji to see if we can publish a release soon)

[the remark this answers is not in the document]
austinEng: Ah, are you saying we have the wrong handedness in all of the samples? If so we should definitely fix that. We should probably also pull in @toji's matrix update and use it everywhere. Right now it's "working" in other places because it doesn't matter significantly, but it would be good to have the examples doing the correct thing for other people to reference.

shrekshao:
Thanks for the review
Another topic is why we need a reversed depth buffer cts test (isn't this a usage case but not verifying webgpu implementation)?
How do you the the cts should look like? My original idea is having two planes placed close to each other and detect the color of their overlap part. But that part is flickering. how do yo reliably detect the flicker?

austinEng:
Thanks, looks good! (do you have the merge button?)
If gl-matrix hasn't published a new version yet, no need to wait for that.

> Another topic is why we need a reversed depth buffer cts test (isn't this a usage case but not verifying webgpu implementation)?
> How do you the the cts should look like? My original idea is having two planes placed close to each other and detect the color of their overlap part. But that part is flickering. how do yo reliably detect the flicker?

Hm I'm not sure what we want for CTS testing and if we really need to test precision. I guess @kainino0x would be the one to ask.

src/pages/samples/reversedZ.ts
```diff
@@ -0,0 +1,1008 @@
...
+fn main() -> void {
+  outColor = fragColor;
+}
+`,
+  vertexDepthPrePass: `
+[[block]] struct Uniforms {
+  [[offset(0)]] modelMatrix : [[stride(${matrixStride})]] array<mat4x4<f32>, ${numInstances}>;
+};
+[[block]] struct Camera {
+  [[offset(0)]] viewProjectionMatrix : mat4x4<f32>;
```
austinEng, line 898:
```suggestion
  viewProjectionMatrix : mat4x4<f32>;
```
offset shouldn't be required anymore

src/pages/samples/reversedZ.ts
```diff
@@ -0,0 +1,1008 @@
...
+[[location(0)]] var<out> outColor : vec4<f32>;
+
+[[stage(fragment)]]
+fn main() -> void {
+  outColor = fragColor;
+}
+`,
+  vertexDepthPrePass: `
+[[block]] struct Uniforms {
+  [[offset(0)]] modelMatrix : [[stride(${matrixStride})]] array<mat4x4<f32>, ${numInstances}>;
```
austinEng, line 895:
```suggestion
  modelMatrix : [[stride(${matrixStride})]] array<mat4x4<f32>, ${numInstances}>;
```
offset shouldn't be required anymore

src/pages/samples/reversedZ.ts
```diff
@@ -0,0 +1,1008 @@
...
+}
+`,
+  fragmentDepthPrePass: `
+[[stage(fragment)]]
+fn main() -> void {
+}
+`,
+  vertexPrecisionErrorPass: `
+[[block]] struct Uniforms {
+  [[offset(0)]] modelMatrix : [[stride(${matrixStride})]] array<mat4x4<f32>, ${numInstances}>;
```
austinEng, line 921:
```suggestion
  modelMatrix : [[stride(${matrixStride})]] array<mat4x4<f32>, ${numInstances}>;
```

src/pages/samples/reversedZ.ts
```diff
@@ -0,0 +1,1008 @@
...
+};
+[[block]] struct Camera {
+  [[offset(0)]] viewProjectionMatrix : mat4x4<f32>;
```
austinEng, line 924:
```suggestion
  viewProjectionMatrix : mat4x4<f32>;
```

src/pages/samples/reversedZ.ts
```diff
@@ -0,0 +1,1008 @@
...
+};
+
+[[builtin(instance_index)]] var<in> instanceIdx : i32;
```
austinEng, line 927:
```suggestion
[[builtin(instance_index)]] var<in> instanceIdx : u32;
```

austinEng (commented):
few more shader comments - might need to check that the shaders still work because the signed->unsigned change might change things

shrekshao:
Fixed shader issues and verify that still works in canary. Merging my self.

shrekshao:
Yes. The reverse-z technique is best for float (not unorm) depth. Also the interpretation would be unique. (compared to depth24plus can be implemented as either depth24unorm or depth32float) Fixed in #88

> On my machine only the top-left rectangle flickers. Is that expected?

On my win10 Quadro 2000M the top 4 on the left all flickers (top left is the most obvious, the fourth rarely flickers). That's a behavior that make sense. But since I now changed `depth24plus` to `depth32float` I still see 4 flickering planes. (32float should be better than 24plus in 0 -1 range?) What are other factors that could cause precision difference here?

kainino0x:
Could be not a precision difference on my machine but an architectural difference that makes flickering less likely. Not really sure why I would see it less (mac intel).

austinEng:
note: Shrek opened #87 about this

PR #86, opened 2021-03-31
Some trivial update and cleanup

shrekshao:
* shader update: some builtin `i32`->`u32`; remove unneeded `[[offset(0)]]`
* replace 36 with cubeDrawCount, should make it a bit easier for replacing other geometry
* a missed style fix for reversedZ


src/pages/samples/fractalCube.ts
```diff
@@ -5,6 +5,7 @@ import {
   cubeColorOffset,
   cubeUVOffset,
   cubePositionOffset,
+  cubeDrawCount,
```
austinEng, line 8: nit: perhaps let's call this `cubeVertexCount` ?

austinEng (approved):

austinEng:
Thanks for the cleanup!

PR #93, opened 2021-04-09
Use new WGSL entry point IO syntax

jrprice:
WGSL now uses entry point function parameters and return values for entry point IO, instead of module-scope variables.

I made some fairly arbitrary choices about when to use structures for input parameters, and on how to name/format things. Please feel free to suggest/make changes as desired.

Tested with latest Chrome Canary on macOS.


austinEng (approved):
Thank you!

PR #94, opened 2021-04-13
Remove unnecessary compute dispatch call in the image blur example.

rvadhavk:
I might be missing something, but I think the `computePass.dispatch(2, Math.ceil(srcHeight / batch[1]));` is just redoing some of the computation from the previous dispatch call.  Removing this line seems to have no effect on the output.


austinEng (approved):
err whoops. I was probably debugging something with that and accidentally left it in. Thanks for the fix!

PR #97, opened 2021-04-18
Use "compute" instead of "computeStage" to silence deprecation warning.

rvadhavk:


austinEng:
Was this closed because we need to update @gpuweb/types for computeStage->compute ?

rvadhavk:
When the build failed, I saw @webgpu/types needed to be updated and made the incorrect assumption that @webgpu/types didn't maintain backward compatibility, so I'd have to do the `view` -> `attachment` rename in the same PR.  I also didn't realize Chrome Canary just added support for the `view` field in GPURenderRenderPassDescriptor, so I thought that rename wouldn't work.

PR #103, opened 2021-04-22
Deferred Rendering

shrekshao:
just yet another sample...


austinEng (commented):
Sorry if this is annoying, but I just pushed a new UI/helpers for computeBoids that has a tab view per source file and separate .wgsl files. If you could update this to be like that, I think it would be great.

shrekshao:
> a tab view per source file and separate .wgsl files.

What do you think would be a good solution for formatting string if read from file? `${x}`?

austinEng:
> > a tab view per source file and separate .wgsl files.
> 
> What do you think would be a good solution for formatting string if read from file? `${x}`?
> 
> string.replace with sth like `${0}`, `${1}`?

Don't have a great solution for that right now :-/ WGSL will eventually have specialization constants in the language which will be the best way to do it.

For now, do you think you can upload those constants to a uniform buffer? I looked briefly at the shaders and I think it could be done.

src/pages/samples/deferredRendering.ts
```diff
@@ -0,0 +1,826 @@
...
+    const mapping = new Uint16Array(indexBuffer.getMappedRange());
+    for (let i = 0; i < mesh.triangles.length; ++i) {
+      mapping.set(mesh.triangles[i], 3 * i);
+    }
+    indexBuffer.unmap();
+  }
+
+  // GBuffer texture render targets
+  // Currently chrome still do not support layered rendering.
+  // Use multiple Texture2D instead of Texture2DArray as render target
```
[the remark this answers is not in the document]
shrekshao: It's this code that make me think it's still not doable. Or is it just validation unremoved but feature is already in?
https://source.chromium.org/chromium/chromium/src/+/master:third_party/dawn/src/dawn_native/CommandEncoder.cpp;l=103?q=%22layered%20rendering%22%20f:dawn&ss=chromium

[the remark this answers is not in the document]
austinEng: if we pass arrayLayerCount: 1 in the view descriptor, `attachment->GetLayerCount()` there *should* return 1 - unless we have some other bug somewhere. Is it returning something else?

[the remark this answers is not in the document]
shrekshao: Haven't tried yet. Was planning for tomorrow (faceplam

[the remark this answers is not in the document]
shrekshao: Yep that just worked.
I was thinking using texture_2d_array<type> in wgsl at very first, so that would end in `attachment->GetLayerCount() > 1`. But anyway that's not very useful since we have differrent formats for g-buffers.

src/sample/deferredRendering/main.ts
```diff
@@ -0,0 +1,616 @@
...
+      ],
+    },
+  ];
+
+  const primitive: GPUPrimitiveState = {
+    topology: 'triangle-list',
+    cullMode: 'back',
+  };
+
+  const WriteGBuffersPipeline = device.createRenderPipeline({
```
austinEng, line 113: nit: variables should be lowercase

```suggestion
  const writeGBuffersPipeline = device.createRenderPipeline({
```

src/sample/deferredRendering/main.ts
```diff
@@ -0,0 +1,616 @@
...
+    usage: GPUBufferUsage.UNIFORM | GPUBufferUsage.COPY_DST,
+  });
+
+  const cameraUniformBuffer = device.createBuffer({
+    size: 4 * 16, // 4x4 matrix
+    usage: GPUBufferUsage.UNIFORM | GPUBufferUsage.COPY_DST,
+  });
+
+  const sceneUniformBindGroup = device.createBindGroup({
+    layout: WriteGBuffersPipeline.getBindGroupLayout(0),
```
austinEng, line 277:
```suggestion
    layout: writeGBuffersPipeline.getBindGroupLayout(0),
```

src/sample/deferredRendering/main.ts
```diff
@@ -0,0 +1,616 @@
...
+      cameraViewProj.byteLength
+    );
+
+    const commandEncoder = device.createCommandEncoder();
+    {
+      // geometry pass write to g buffers
+      const gBufferPass = commandEncoder.beginRenderPass(
+        writeGBufferPassDescriptor
+      );
+      gBufferPass.setPipeline(WriteGBuffersPipeline);
```
austinEng, line 519:
```suggestion
      gBufferPass.setPipeline(writeGBuffersPipeline);
```

src/sample/deferredRendering/main.ts
```diff
@@ -0,0 +1,616 @@
...
+      entryPoint: 'main',
+      buffers: vertexBuffers,
+    },
+    fragment: {
+      module: device.createShaderModule({
+        code: fragmentWriteGBuffers,
+      }),
+      entryPoint: 'main',
+      targets: [
+        { format: 'rgba32float' },
```
austinEng, line 127: nit: probably comment what each of these in the gBuffer stores

src/sample/deferredRendering/fragmentDeferredRendering.wgsl
```diff
@@ -0,0 +1,69 @@
...
+  radius : f32;
+};
+[[block]] struct LightsBuffer {
+  lights: array<LightData>;
+};
+[[group(1), binding(0)]] var<storage> lightsBuffer: [[access(read)]] LightsBuffer;
+
+[[block]] struct Config {
+  numLights : u32;
+};
```
austinEng, line 18: optional, but if your lights buffer is of fixed length, then you could use `arrayLength(lightsBuffer.lights)` in the shader instead

src/sample/deferredRendering/main.ts
```diff
@@ -0,0 +1,616 @@
...
+    requestAnimationFrame(frame);
+  }
+  requestAnimationFrame(frame);
+};
+
+const DeferredRendering: () => JSX.Element = () =>
+  makeSample({
+    name: 'Deferred Rendering',
+    description:
+      'This example shows how to do deferred rendering with webgpu. Render geometry info to GBuffers and do the lighting independent of scene complexity.',
```
austinEng, line 574: maybe have a brief description as well that we write the position, uvs, etc. to multiple render targets and then compute the lighting results in a separate pass.

austinEng (approved):
Looks good - though we could probably have more comments so that people reading the code get a better understanding. That said, I guess we're not attempting to teach deferred rendering in this demo

src/sample/deferredRendering/fragmentDeferredRendering.wgsl
shrekshao: New things learnt. I also learned that we can do non-const for loop range in wgsl.

shrekshao:
Thanks for the review

austinEng:
Still LGTM, feel free to merge if you'd like

PR #108, opened 2021-04-28
Add sample for resizing canvas

austinEng:


shrekshao (commented):
When switching away from resizingCanvas to other samples there's
```
Unhandled Runtime Error
TypeError: Cannot read property 'clientWidth' of null
```

seems `frame` is still called but canvasRef in that scope is lost. Somthing related to the react framework?

austinEng:
> When switching away from resizingCanvas to other samples there's
> 
> ```
> Unhandled Runtime Error
> TypeError: Cannot read property 'clientWidth' of null
> ```
> 
> seems `frame` is still called but canvasRef in that scope is lost. Somthing related to the react framework?

Good catch - I think this was actually a problem with all the new ui samples. I pushed a change that should fix it.

shrekshao (approved):
LGTM!

PR #112, opened 2021-05-27
Fix shadow mapping diffuse component to use max function

liukun6:
Use the max function to make sure the diffuse component never becomes negative.


austinEng (approved):
thanks for the fix!

PR #115, opened 2021-06-07
wgsl: Remove `-> void`

ben-clayton:
`void` was removed in https://github.com/gpuweb/gpuweb/pull/1460


austinEng:
I think you need a rebase?

ben-clayton:
My git `origin` was pointing at the wrong thing, which was ancient. Ignore me.

PR #116, opened 2021-06-11
Updates for WGSL spec changes

ben-clayton:
Remove access control decorations.
See https://github.com/gpuweb/gpuweb/pull/1735

arrayLength() now takes a pointer to the array.
See https://github.com/gpuweb/gpuweb/pull/1619


austinEng (approved):
Thanks!

PR #117, opened 2021-06-17
wgsl: Fixes for new validation failures

ben-clayton:
Values of array and matrix can now only be indexed by a constant value.
Dynamic indexing requires the value to be held in memory.

See: gpuweb/gpuweb#1782


Kangz (approved):

Kangz:
Ah I don't have write access to this repo. @kainino0x can you help?

austinEng (approved):

Issue #118, opened 2021-06-17
Take advantage of `var` type omission in WGSL


Kangz:
This would make the shaders nicer to read!

alyian:
![IMG_٢٠٢١٠٧٠٩_٠٢٤٥٢١](https://user-images.githubusercontent.com/86137533/125208730-006fc580-e29d-11eb-861c-f56b48dfb468.jpg)

alyian:
![IMG_٢٠٢١٠٧٠٨_١٤٢٦٢٢](https://user-images.githubusercontent.com/86137533/125208752-15e4ef80-e29d-11eb-9ccc-9ee726124d82.jpg)

alyian:
![IMG_٢٠٢١٠٦٠٦_٠٠٥٢٢٨](https://user-images.githubusercontent.com/86137533/125208758-26956580-e29d-11eb-8d5e-f1d467577841.png)

alyian:
![IMG_٢٠٢١٠٦١٦_١١١٤٢٧](https://user-images.githubusercontent.com/86137533/125208800-5a708b00-e29d-11eb-9a48-55c89d077ed6.png)

PR #121, opened 2021-06-24
Take advantage of var type

munrocket:
Fixes: #118
All examples looks the same in Chrome Canary.


kainino0x (approved):
LGTM!

I bet many of these could be `let`, but that can be a change for another time :)

PR #122, opened 2021-06-29
wgsl: Add workgroup_size() decoration on compute shader

ben-clayton:
This is now required by validation.

16 is a number I made up. It is larger than 1, which was the old default.

Note: I haven't actually checked that `workgroup_size(16)` actually behaves correctly.


austinEng:
Probably need to update the dispatch command to divide by 16, then ceiling. Feel free to just change this to 1 though to keep the old behavior, and perhaps @shrekshao can look at making the workgroup size more efficient.

ben-clayton:
LGTM if you want to land

austinEng (approved):

PR #129, opened 2021-07-28
Fix deferredRendering and imageBlur samples

austinEng:


shrekshao (approved):
This is fast. Thanks.
(learnt: textureLoad for unfilterable formats)

austinEng:
> (learnt: textureLoad for unfilterable formats)

can also use a sampler, but you need to specify 'nearest' for magFilter, minFilter, mipmapFilter

Issue #180, opened 2022-05-12
Visible artifacts in imageBlur 


jereroy:
There are visible artifacts in the image blur sample. With 10 iterations and a filter size of 2, they are particularly obvious :

![blurArtifact](https://user-images.githubusercontent.com/25518631/167977311-20dbecd2-333c-4166-a306-e285039aacf3.png)

Also, I feel like the goal was to have an odd size filter but dat.gui rounds the size to an even number atm.

PR #181, opened 2022-05-12
Fix blur artifacts with odd filter size

jereroy:
See #180

### DIFF
--- a/src/sample/imageBlur/main.ts
+++ b/src/sample/imageBlur/main.ts
@@ -222,6 +222,8 @@ const init: SampleInit = async ({ canvasRef, gui }) => {
 
   let blockDim: number;
   const updateSettings = () => {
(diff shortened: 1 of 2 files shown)


src/sample/imageBlur/main.ts
```diff
@@ -222,6 +222,8 @@ const init: SampleInit = async ({ canvasRef, gui }) => {
 
   let blockDim: number;
   const updateSettings = () => {
+    settings.filterSize += 1;
```
austinEng, line 225: instead of += 1, I think it would be better to change the computations below so that they are `settings.filterSize` and `settings.filterSize + 1` instead of `settings.filterSize - 1` and `settings.filterSize`

austinEng (commented):
thank you for the contribution! sorry it took a while for me to review this

PR #187, opened 2022-06-03
Fix resizeCanvas sample to use devicePixelRatio

### LINKED ISSUE #176: ResizeCanvas sample doesn't properly handle devicePixelRatio
As far as I can tell, `devicePixelRatio` is handled in the original declaration of `presentationSize`, but not in testing the size for changes, or in setting the size after changes.

austinEng:
Fixes #176


shrekshao (approved):
LGTM

PR #191, opened 2022-07-08
Update WGSL

onsummer:
As the title.

## WGSL update

- Using `@vertex`, `@fragment`, `@compute` instead of `@stage(vertex)`, `@stage(fragment)` and `@stage(compute)`
- Change the `;` at the end of the member field in the struct declaration to `,`
- Remove the unnecessary `;` at the end of struct declaration
- Using compound assignment
- Beautify WGSL code indentation

And I added an icon for this pages.


onsummer:
@austinEng I run `lint` script with my Windows 11 and node 16.x, there are no errors and warns

image

Really don't know why it build failure, do you have ideas? 😂

onsummer:
Because 16.x is a lts version? Is it the main reason? If so I change it back to 15.x. 

Before this, there is a missing type in `navigation.gpu`:

```
./src/sample/imageBlur/main.ts:17:44
Type error: Property 'getPreferredCanvasFormat' does not exist on type 'GPU'.
 15 | const context = canvasRef.current.getContext('webgpu') as GPUCanvasContext;
 16 | 
> 17 | const presentationFormat = navigator.gpu.getPreferredCanvasFormat();
```

austinEng:
Ok, makes sense to update to 16.x since it's LTS.
For this PR, I think we need to update the version of @webgpu/types since that provides the types for `GPU` and `getPreferredCanvasFormat`

onsummer:
OK, I have been updated `@webgpu/types` to latest version. Since this package is important for type hinting for the WebGPU api, I explicitly set it to latest in package.json.

onsummer:
Hope this time will success

austinEng (approved):
LGTM!

PR #192, opened 2022-07-08
Update WebGPU API and formatting

onsummer:
## WebGPU API update

- Change `context.getPreferredFormat(adapter)` to `navigator.gpu.getPreferredCanvasFormat()`
- Remove the deprecated `size` in `GPUCanvasConfiguration`, and explicit set `alphaMode` in `GPUCanvasConfiguration`
- Add the required `layout` field for `GPUPipelineDescriptorBase` (both `GPURenderPipeline` and `GPUComputePipeline`)
- Change `GPUComputePassEncoder.dispatch` to `GPUComputePassEncoder.dispatchWorkgroups` during passes encoding

## Code Formatting

Ran `lint --fix`


onsummer:
@austinEng When my screen display ratio greater than 1 would log this message:

image

My laptop's devicePixelRatio is 1.75

.prettierrc.js
```diff
@@ -1,4 +1,8 @@
 module.exports = {
+  printWidth: 80,
+  tabWidth: 2,
+  useTabs: false,
   semi: true,
   singleQuote: true,
+  endOfLine: "auto",
```
austinEng, line 7: hm, I think we want to be using "lf" which is already the default (so this line is no longer needed).

Looks like this PR has changed multiple files from LF -> CRLF

austinEng:
> For eg., in msaa sample, My laptop's devicePixelRatio is 1.75, so the resolveTexture's size will set to be [600×1.75, 600×1.75] = [1050, 1050], but canvas texture is configure by canvas size, which is [600, 600], and I do not know how to set canvas texture's size, because `size` in `GPUCanvasConfiguration` was marked `@deprecated`.

Hm, I'm asking about this - it could be a bug in the Chrome implementation. For now, could we keep passing the explicit `presentationSize` to `GPUCanvasContext.configure(..)` ? I think that will resolve the problem, but you'll have warnings about it in the console.

.prettierrc.js
onsummer: I use Windows, maybe the reason. Should I change back to LF?

austinEng: Yea, for projects used on multiple OSes, LF is usually the preferred line-ending. I think if you remove this line and then run `npm run lint -- --fix` it'll fix all of them for you.

onsummer: OK, later

onsummer:
Learn a lot :)

PR #194, opened 2022-07-10
Update canvas size configuration when devicePixelRatio != 1

onsummer:
- using CSS to control canvas DOM's height and width
- multiply devicePixelRatio to canvas' height and width
- remove `size` field when `GPUCanvasContext.configure()`

All samples work well, the console is clean again


onsummer:
@austinEng PTAL

src/sample/resizeCanvas/main.ts
```diff
@@ -56,56 +51,53 @@ const init: SampleInit = async ({ canvasRef }) => {
...
-    const currentHeight = canvasRef.current.clientHeight * devicePixelRatio;
+    if (!canvas) return;
 
-    // The canvas size is animating via CSS.
-    // When the size changes, we need to reallocate the render target.
-    // We also need to set the physical size of the canvas to match the computed CSS size.
+    const devicePixelRatio = window.devicePixelRatio;
+    // Make sure canvas size to be integer.
+    const currentWidth = Math.floor(canvas.clientWidth * devicePixelRatio);
+    const currentHeight = Math.floor(canvas.clientHeight * devicePixelRatio);
```
austinEng, line 77: A few simplications we can do here, and some optimizations:

- I don't think we need `presentationSize`. We can use `canvas.width` directly.
- The existing code shouldn't need to check the canvas height since it's not animating. We can remove that and only use the width.
- I think we should "bucket" the canvas size so that sometimes it's slightly larger than necessary. This will let us avoid reallocating the canvas backing textures more than they need to be and give a smoother animation. Something like `const width = Math.ceil(canvas.clientWidth * devicePixelRatio / 50) * 50;`

src/components/SampleLayout.tsx
```diff
@@ -102,6 +102,15 @@ const SampleLayout: React.FunctionComponent<
...
 
+  // Ref it because it is a static number,
+  // no need to be seen as a React state during virtual dom update.
+  const devicePixelRatioRef = useRef<number>(1);
+  // And we have to cache devicePixelRatio when component did mount,
+  // because `window` can not be accessed except this hook. -- NextJS
+  useEffect(() => {
+    devicePixelRatioRef.current = window.devicePixelRatio;
+  }, []);
+
```
austinEng, line 113: would it be sufficient to just do:

`const dpr = process.browser ? window.devicePixelRatio : 1;`?

process.browser should make it so that in the static build when we don't have `window`, we use 1, but when the component mounts, we'll look at `window.devicePixelRatio`

src/components/SampleLayout.module.css
```diff
@@ -4,6 +4,11 @@
   margin-top: 10px;
 }
 
+.canvasContainer canvas:first-of-type {
```
austinEng, line 7: nit: instead of using first-of-type, I think it would be better to give the canvas an id attribute in the HTML, and use #id here.

src/components/SampleLayout.tsx
onsummer: Take a look:

```jsx
const SampleLayout = (props) => {
  /* ... */
  const dpr = process.browser ? window.devicePixelRatio : 1;
  /* ... */

  return (
    /* ... */
    <canvas
      ref={canvasRef}
      id={styles.renderTarget}
      width={600 * dpr}
      height={600 * dpr}
    ></canvas>
    /* ... */
  )
}
```

image

src/sample/resizeCanvas/main.ts
austinEng: The width is animating with CSS from 10px -> 600px.

For simplicity, let's say `dpr == 2`.
I think the actual `canvas.width` should get set such that:
 CSS width | Canvas width 
----------- | -------------
10-50         | 100
51 - 100     | 200
...                | ...
551 - 600  | 1200

This way, we're not _constantly_ updating `canvas.width` which will cause a new texture to be allocated.

src/components/SampleLayout.tsx
austinEng: NextJS still renders a static page so that the initial .html isn't blank.
This warning is happening because we're changing the size to a different value - which is quite intentional but React is upset.
What if instead we set width/height entirely outside of React props so they're not tracked by React at all?

For example, using a ref callback:
```tsx
        <canvas
          ref={(canvas) => {
            canvasRef.current = canvas;
            if (canvas) {
              canvas.width = 600 * window.devicePixelRatio;
              canvas.height = 600 * window.devicePixelRatio;
            }
          }}
        />
```

src/sample/resizeCanvas/main.ts
onsummer: Great. I set the canvas texture reallocated interval to 200px, and in a 15-second chrome performance test, the createTexture function was called only 11 times. 

If the interval is 50, then will be nearly 60 calls.

src/components/SampleLayout.tsx
onsummer: The example works!

src/sample/resizeCanvas/main.ts
austinEng: Hm, I don't think this is quite what we want. As you have it now, as the canvas is getting wider, it will be smaller than we want it. For example, consider:

- `canvas.clientWidth` is currently 10px, `currentWidthStep` is ceil(10/200) == 1, we update `canvas.width` to 10px * dpr.

 - Now, the canvas animates, and for example is 200px, `currentWidthStep` is ceil(10/200) == 1. `currentWidthStep` still equals `lastWidthStep`, so `canvas.width` is not updated. It is still 10px * dpr when the actual CSS width is 200.

The problem is that we're always updating the canvas size to be the lower end of the 200px "bucket".
Instead of setting canvas.width to be `currentWidth`, I think we should set it to be `ceil(currentWidthStep * 200 * dpr)` which would be the higher end of the 200px "bucket".

austinEng:
I think @toji has already fixed these issues, so closing this PR.

PR #203, opened 2022-09-25
Update origin-trial

puxiao:
The previous `origin-trial` has expired, I reapplied for one.

Currently chrome 105+ can run WebGPU.


austinEng:
Ah, I went and updated the token myself before I had seen this. Thank you for making the PR.

PR #220, opened 2023-01-23
Fix `textureSampleLevel` for external texture

ben-clayton:
This now needs to use `textureSampleBaseClampToEdge()`


austinEng (approved):
Thanks!

Issue #226, opened 2023-02-02
Transferring repository to https://github.com/webgpu/samples


austinEng:
We'd like to transfer this repository to https://github.com/webgpu/samples so it's in a shared community organization and not under my personal github account :)

Creating this issue to seek any feedback/comments from the various non-Google contributors. Thank you for your contributions!
 - @onsummer 
 - @munrocket 
 - @HypnosNova
 - @webglzhang 
 - @inokawa
 - @rvadhavk
 - @lukehesluke
 - @frguthmann
 - @ponitka
 
I'll wait for a period of about two weeks, and if there are no strong objections, I'll transfer the repo on February 17, 2023.

HypnosNova:
That's a great idea.
And my pr: https://github.com/austinEng/webgpu-samples/pull/204 
I think this is a good sample but no one cr. It would be better merge into master before transfer repository to webgpu/samples

hepp:
This is great news!
Thank you for this, I'm looking forward to seeing a thriving platform for contributions.

If interested - I have created several new examples at https://www.wgsl.dev/, which could be ported over.

Also below are a couple ideas on what I might be working on next / or what we could potentially add here.

- Example of Video 3D Plane
- Example of Quad raymarching 
- Example of Light with a simplified deferred setup

austinEng:
The transfer is complete!

Issue #82, opened 2021-03-11
Text rendering example (that uses canvas as texture)


smolck:
This may be a bit much to ask (in which case no worries, I totally understand), but I figure I'll ask anyways; I think it would be quite useful to have an example of rendering text, specifically one that first renders the text to a canvas and then renders that with WebGPU via a texture. 

A more complicated but also perhaps more useful example would render some chars to a canvas, create a texture from it, and then use that texture to construct various strings on-the-fly by sampling from different parts of the texture for each letter. (Note that a WebGL version of this is described in https://webglfundamentals.org/webgl/lessons/webgl-text-glyphs.html, which might be helpful.)

I'm happy to help with the implementation of this if wanted and as I'm able, but unfortunately I think my knowledge is too lacking in both WebGPU and graphics programming in general to do the full PR myself.

kainino0x:
This would be a super valuable sample! I definitely think we should have it.

However the APIs for getting canvas data into WebGPU are only just starting to take form.
https://github.com/gpuweb/gpuweb/issues/1154
https://github.com/gpuweb/gpuweb/issues/1415#issuecomment-796880809
I suspect the simpler one (`copyToTexture`) would be ideal for building glyph caches/atlases, while GPUExternalTexture would be better for rendering an entire text block that changes every frame.

smolck:
> However the APIs for getting canvas data into WebGPU are only just starting to take form.

Interesting, so does that mean WebGPU doesn't support canvases as textures yet, or just that it's not ergonomic? I've tried to get the canvas into a texture using something like this:
```javascript
  const texture = device.createTexture({
    size: [canvas.width, canvas.height, 1],
    format: 'rgba8unorm',
    usage: GPUTextureUsage.SAMPLED | GPUTextureUsage.COPY_DST,
  })
  const ctx = canvas.getContext('2d')
  const imageData = ctx.getImageData(0, 0, canvas.width, canvas.height)
  const canvasBitmap = await createImageBitmap(imageData)
  device.queue.copyImageBitmapToTexture(
    { imageBitmap: colorAtlasImageBitmap, },
    { texture: colorAtlasTexture },
    [colorAtlasImageBitmap.width, colorAtlasImageBitmap.height, 1]
  )
```
That doesn't seem to work though?

kainino0x:
It's not possible without some indirection through ImageBitmap or writeTexture just like what you've done.
[...]

Issue #45, opened 2020-07-29
Add a mipmap generation example.


Kangz:
For example based on this [code](https://gist.github.com/Kangz/782d5f1ae502daf53910a13f55db2f83).

austinEng:
I also should actually update things to use `copyImageBitmapToTexture` instead of the CPU readback and B2T copy. @kvark is `copyImageBitmapToTexture` implemented in Firefox or will I need to polyfill with `writeTexture`?

kvark:
`writeTexture` is in, `copyImageBitmapToTexture` is not

munrocket:
Example is added in #45

Issue #41, opened 2020-05-14
Web Workers


ottoville:
According to WEBGPU spec, WEBGPU objects such as buffers should be able to move between worker threads.
Should there be example for doing that?

kainino0x:
AFAIK there are no implementations of that part of the spec yet. (Chrome does not implement it.) It's still in flux because the design hasn't been proven out yet. I'm sure a sample will appear once someone starts implementing it.

kainino0x:
This is not in the spec anymore. We'll eventually add a sample whenever it gets added to the spec.

PR #227, opened 2023-02-07
Boids Sprites: rotation and velocity behavior visualized with color

hepp:
Thanks for putting together these great samples!

Here is a small addition to help improve visualizing boids (subjective).
Solely aesthetic - and this change could be viewed subjectively if it’s an improvement or not.

To me, this greatly helps in visualizing the dense bird movements, making the cluster and movements appear more 3D.
Thought I would share, after implementing this here: https://www.wgsl.dev/editor?example=particles&canvas=webgpu
No worries if you don’t think it’s necessary to include here.

### DIFF
--- a/src/sample/computeBoids/sprite.wgsl
+++ b/src/sample/computeBoids/sprite.wgsl
@@ -25,10 +25,6 @@ fn vert_main(
   return output;
 }
 
-struct FragmentInput {
-  @location(4) color : vec3<f32>,
-}
-
 @fragment
 fn frag_main(@location(4) color : vec4<f32>) -> @location(0) vec4<f32> {
   return color;
--- a/src/sample/computeBoids/sprite.wgsl
+++ b/src/sample/computeBoids/sprite.wgsl
@@ -1,18 +1,35 @@
+struct VertexOutput {
+  @builtin(position) position : vec4<f32>,
+  @location(4) color : vec4<f32>,
+}
+
 @vertex
 fn vert_main(
   @location(0) a_particlePos : vec2<f32>,
   @location(1) a_particleVel : vec2<f32>,
   @location(2) a_pos : vec2<f32>
-) -> @builtin(position) vec4<f32> {
+) -> VertexOutput {
   let angle = -atan2(a_particleVel.x, a_particleVel.y);
   let pos = vec2(
     (a_pos.x * cos(angle)) - (a_pos.y * sin(angle)),
     (a_pos.x * sin(angle)) + (a_pos.y * cos(angle))
   );
-  return vec4(pos + a_particlePos, 0.0, 1.0);
+  
+  var output : VertexOutput;
+  output.position = vec4(pos + a_particlePos, 0.0, 1.0);
+  output.color = vec4(
+    1.0 - sin(angle + 1.0) - a_particleVel.y,
+    pos.x * 100.0 - a_particleVel.y + 0.1,
+    a_particleVel.x + cos(angle + 0.5),
+    1.0);
+  return output;
+}
+
+struct FragmentInput {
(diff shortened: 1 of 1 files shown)


austinEng:
checking: are you aware of #226 and that this repository will be transferred to a new org?

hepp:
I wasn't aware. Looks like good news. 
However you'd like to proceed.

I've made several improvements that might be of interest.
I could just wait on those until the transfer is completed.

src/sample/computeBoids/sprite.wgsl
```diff
@@ -1,18 +1,35 @@
...
+  return output;
 }
 
-@fragment
-fn frag_main() -> @location(0) vec4<f32> {
-  return vec4(1.0, 1.0, 1.0, 1.0);
+struct FragmentInput {
+  @location(4) color : vec3<f32>,
 }
+
```
austinEng, line 31: this struct is unused since you're passing the color in as `@location(4) color : vec4<f32`

hepp: Good catch, thank you.

austinEng (approved):

austinEng:
> I've made several improvements that might be of interest.

Feel free to make PRs. There isn't a real difference between whether we do it now or after the transfer.

hepp:
Sounds good. I certainly will have some more coming shortly.
Currently a bit busy, but once my current contract completes in a couple months I'll have lot more time again.

I'm curious - 
Is there any interest in expanding the amount of examples?
Versus improving upon the current ones?

austinEng:
More examples can be good if they teach new concepts. For example there are various issues in this repo asking for
 - mipmap generation #45 
 - text rendering #82
 - web workers #41

We'll need to figure out how structuring the information will be most useful in the community.
currently the samples are a mix of super basic stuff - and then it jumps to "here's a cool demo!". Maybe that's fine, maybe not. Overall, I think we could have:
 - some partition between "basics" and "demos", where the basics help showcase simple components used to make bigger demos
 - more text to explain what is going on. Maybe this is better suited for a "WebGPU book"
 - fix the various issues about the code editor being broken
 - restructure things to use iframes
    - for some reason, I didn't want to use iframes in the past, but it would be admittedly much simpler than the way I'm extracting the code to display on the page
    - it would also make reloading the iframe on shader changes for live editing (and we can make the js editable too I suppose :o). I used to have some mechanism to do shader editing which is now broken after some spec changes

hepp:
Ok, yes that is a great start of a list of new examples to aim for.

Also in response to the points you mentions in how the samples are structured.

> - some partition between "basics" and "demos", where the basics help showcase simple components used to make bigger demos

Agreed, particularity examples utilizing compute shaders and deferred rendering techniques.

> - more text to explain what is going on. Maybe this is better suited for a "WebGPU book"

That could certainly be helpful.
There is probably a better way to include that without lots of extraneous comments in the code. 
 
The way that Svelte handles it's [tutorial](https://svelte.dev/tutorial/basics), has always seemed like a good format for including explanations alongside code samples.

> - fix the various issues about the code editor being broken

Yes it appears current issues with the code editor might revolve around incorrectly nested dom elements, which are causing miss layering during vertical scrolling.
The elements that are assigned `overflow-y` class styles might need to be revised.

Also it might be helpful to enable the ability to edit the main file of each example,
(which doesn't seem like should directly include TypeScript or JSX code).

> - restructure things to use iframes

Working with iframes so far has seemed to work out great.
Using Node server with React SSR, seems to result in nice performance between editor and iframe. 

Injecting in additional meta and script tags, is one [solution](https://github.com/hepp/webgpu-examples#in-addition) I came up with, for keeping extra code out of the examples in order to keep them short and succinct to the example demonstrated.

Issue #237, opened 2023-03-17
The deferred renderer example has a validation error with maxColorAttachmentBytesPerSample.


Kangz:
```
Total color attachment bytes per sample (40) exceeds maximum (32) with formats ([ TextureFormat::RGBA32Float TextureFormat::RGBA32Float TextureFormat::BGRA8Unorm ]).
 - While validating fragment state.
 - While calling [Device].CreateRenderPipeline([RenderPipelineDescriptor]).
```

823984418:
This seems to be related to https://github.com/gpuweb/gpuweb/pull/3367 .

But I never seem to be able to request a higher limit.

shrekshao:
If not able to request a higher limit, to fix the sample, I think one easier change is to use 'float16x4' and 'RGBA16Float' for the second render target (normal)

(32 bytes per sample for color attachment seems to be quite strict for any float target then)

PR #238, opened 2023-03-21
[Deferred Rendering] Use 'rgba16float' format for normal gbuffer

shrekshao:
Use rgba16float texture instead of rgba32float for normal gbuffer to avoid hitting maxColorAttachmentBytesPerSample limit.
Fixing https://github.com/webgpu/webgpu-samples/issues/237 to make deferred rendering sample work again


toji (approved):
Awesome, thank you! I was going to try and fix this today in order to get some screenshots of it, so I really appreciate this!

PR #246, opened 2023-04-07
Fix Fractal Cube sample name

akd-io:
Was incorrectly identical to TexturedCube, probably as the sample was copy-pasted


austinEng (approved):

austinEng:
Thank you!

PR #247, opened 2023-04-07
Add Video Uploading with WebCodecs sample

beaufortfrancois:
This PR adds a sample "Video Uploading with WebCodecs", similar to ["Video Uploading"](https://webgpu.github.io/webgpu-samples/samples/videoUploading), so that web developers can play with  `importExternalTexture()` and `VideoFrame` from Web Codecs API.

image

### DIFF
--- a/src/sample/videoUploadingWebCodecs/main.ts
+++ b/src/sample/videoUploadingWebCodecs/main.ts
@@ -142,9 +142,9 @@ const init: SampleInit = async ({ canvas, pageState }) => {
   }
 };
 
-const VideoUploading: () => JSX.Element = () =>
+const VideoUploadingWebCodecs: () => JSX.Element = () =>
   makeSample({
-    name: 'Video Uploading with WebCodecs',
+    name: 'Video Uploading with WebCodecs (Experimental)',
     description: 'This example shows how to upload VideoFrame to WebGPU.',
     init,
     sources: [
@@ -166,4 +166,4 @@ const VideoUploading: () => JSX.Element = () =>
     filename: __filename,
   });
 
-export default VideoUploading;
+export default VideoUploadingWebCodecs;
--- a/src/sample/videoUploadingWebCodecs/main.ts
+++ b/src/sample/videoUploadingWebCodecs/main.ts
@@ -101,7 +101,7 @@ const init: SampleInit = async ({ canvas, pageState }) => {
         {
           binding: 2,
           resource: device.importExternalTexture({
-            source: videoFrame as HTMLVideoElement,
+            source: videoFrame as any, // eslint-disable-line @typescript-eslint/no-explicit-any
           }),
         },
       ],
@@ -145,7 +145,7 @@ const init: SampleInit = async ({ canvas, pageState }) => {
 const VideoUploading: () => JSX.Element = () =>
   makeSample({
     name: 'Video Uploading with WebCodecs',
-    description: 'This example shows how to upload video frame from WebCodecs to WebGPU.',
+    description: 'This example shows how to upload VideoFrame to WebGPU.',
     init,
     sources: [
       {
--- a/src/sample/videoUploadingWebCodecs/main.ts
+++ b/src/sample/videoUploadingWebCodecs/main.ts
@@ -145,7 +145,14 @@ const init: SampleInit = async ({ canvas, pageState }) => {
 const VideoUploadingWebCodecs: () => JSX.Element = () =>
   makeSample({
     name: 'Video Uploading with WebCodecs (Experimental)',
-    description: 'This example shows how to upload VideoFrame to WebGPU.',
+    description: `This example shows how to upload a WebCodecs VideoFrame to WebGPU.
+      Support for using a VideoFrame as the source for a GPUExternalTexture requires
+      running Chrome with the "WebGPU Developer Features" flag or the WebGPU WebCodecs
+      integration origin trial.
+      See https://developer.chrome.com/origintrials/#/view_trial/1705738358866575361
+    `,
+    originTrial:
+      'Auo9JMDbdn/Jg1pd8liB9Ofp1OLzi9mecxjBBfjv/3f8O8775CXgcTobX4t6KYxMC1wnO4Z7MWArPSptGtkD2woAAABZeyJvcmlnaW4iOiJodHRwczovL3dlYmdwdS5naXRodWIuaW86NDQzIiwiZmVhdHVyZSI6IldlYkdQVVdlYkNvZGVjcyIsImV4cGlyeSI6MTcwMTk5MzU5OX0=',
     init,
     sources: [
       {
--- a/src/sample/videoUploadingWebCodecs/main.ts
+++ b/src/sample/videoUploadingWebCodecs/main.ts
@@ -1,0 +1,169 @@
+import { makeSample, SampleInit } from '../../components/SampleLayout';
+
+import fullscreenTexturedQuadWGSL from '../../shaders/fullscreenTexturedQuad.wgsl';
+import sampleExternalTextureWGSL from '../../shaders/sampleExternalTexture.frag.wgsl';
+
+const init: SampleInit = async ({ canvas, pageState }) => {
+  // Set video element
+  const video = document.createElement('video');
+  video.loop = true;
+  video.autoplay = true;
+  video.muted = true;
+  video.src = new URL(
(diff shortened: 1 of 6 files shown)


src/sample/videoUploadingWebCodecs/main.ts
```diff
@@ -0,0 +1,169 @@
...
+      layout: pipeline.getBindGroupLayout(0),
+      entries: [
+        {
+          binding: 1,
+          resource: sampler,
+        },
+        {
+          binding: 2,
+          resource: device.importExternalTexture({
+            source: videoFrame as HTMLVideoElement,
```
beaufortfrancois, line 104: I'd like to import videoFrame `as VideoFrame` but I fail to do that. Any idea how to proceed?

src/types.d.ts
```diff
@@ -7,6 +7,10 @@ declare module '*.module.css' {
 interface HTMLCanvasElement extends HTMLElement {
   getContext(contextId: 'webgpu'): GPUPresentationContext | null;
 }
+interface GPUExternalTextureDescriptor extends GPUObjectDescriptorBase {
+  source: HTMLVideoElement | VideoFrame;
```
beaufortfrancois, line 11: This seems unused ;(

src/sample/videoUploadingWebCodecs/main.ts
kainino0x: Right now the types don't allow it because the spec doesn't allow it. We should probably just introduce a manual diff to the types to allow it. In the meantime this is fine (or `as any`).

src/types.d.ts
kainino0x: The definition of importExternalTexture in `@webgpu/types` will be using the definition of GPUExternalTextureDescriptor from the same file, and not picking this one up. If you also add a definition for importExternalTexture here it might work.

src/sample/videoUploadingWebCodecs/main.ts
kainino0x: Oh, adding it upstream would require a dependency on `@types/dom-webcodecs`. Probably fine but heavier weight than I realized.

beaufortfrancois: I used `as any` for now.

src/types.d.ts
beaufortfrancois: I removed this as I'm using `as any` for now.

beaufortfrancois:
@kainino0x Please have another look. This PR is now ready for review.

src/sample/videoUploadingWebCodecs/main.ts
```diff
@@ -0,0 +1,169 @@
...
+    video.requestVideoFrameCallback(frame);
+  } else {
+    requestAnimationFrame(frame);
+  }
+};
+
+const VideoUploadingWebCodecs: () => JSX.Element = () =>
+  makeSample({
+    name: 'Video Uploading with WebCodecs (Experimental)',
+    description: 'This example shows how to upload VideoFrame to WebGPU.',
```
kainino0x, line 148: I think we should document somewhere in here that an origin trial token is currently required for this in Chrome.

Also do we (/should we) have an origin trial token for this sample?

src/sample/videoUploadingWebCodecs/main.ts
```diff
@@ -0,0 +1,169 @@
...
+
+  const adapter = await navigator.gpu.requestAdapter();
+  const device = await adapter.requestDevice();
+
+  if (!pageState.active) return;
+
+  const context = canvas.getContext('webgpu') as GPUCanvasContext;
+  const devicePixelRatio = window.devicePixelRatio || 1;
+  canvas.width = canvas.clientWidth * devicePixelRatio;
+  canvas.height = canvas.clientHeight * devicePixelRatio;
```
kainino0x, line 26: @austinEng do we have common code anywhere for creating hidpi canvases? we should, and it should use ResizeObserver to handle devicePixelRatio changes (and forward them to the samples)

src/sample/videoUploadingWebCodecs/main.ts
```diff
@@ -0,0 +1,169 @@
...
+  video.src = new URL(
+    '../../../assets/video/pano.webm',
+    import.meta.url
+  ).toString();
+  await video.play();
+
+  const adapter = await navigator.gpu.requestAdapter();
+  const device = await adapter.requestDevice();
+
+  if (!pageState.active) return;
```
kainino0x, line 21: I don't know what this does in the init function, exactly, but seems like it should be before playing the video? (it seems fine if it's before requestAdapter too)

src/sample/videoUploadingWebCodecs/main.ts
```diff
@@ -0,0 +1,169 @@
...
+      topology: 'triangle-list',
+    },
+  });
+
+  const sampler = device.createSampler({
+    magFilter: 'linear',
+    minFilter: 'linear',
+  });
+
+  function getVideoFrameFromVideoElement(video) {
```
kainino0x, line 64: I don't know this API at all so can't really review this. Seem reasonable.

kainino0x (approved):

src/sample/videoUploadingWebCodecs/main.ts
austinEng: I should eventually delete this and use iframes that simply get deleted - but right now the page can get switched out between when the component "mounts" and when the awaits finish. So we would then be rendering into a canvas that's not supposed to be visible anymore.

kainino0x: oh gotcha, so this is correct then (everything before it is an await)

src/sample/videoUploadingWebCodecs/main.ts
beaufortfrancois: I've filed https://github.com/webgpu/webgpu-samples/issues/249 to track this

src/sample/videoUploadingWebCodecs/main.ts
beaufortfrancois: FYI I used similar code to https://github.com/gpuweb/cts/blob/0ac5d46/src/webgpu/web_platform/util.ts#L223

src/sample/videoUploadingWebCodecs/main.ts
beaufortfrancois: Good feedback! I've updated description and added a token for this sample: https://github.com/webgpu/webgpu-samples/pull/247/commits/606cec232c9f4ce1f0e5fc9668e58cb8ad68fcc1

src/sample/videoUploadingWebCodecs/main.ts
kainino0x: It does seem rather complex given that VideoFrame has a constructor that takes an HTMLVideoElement. But maybe this is necessary. It would be nice to document why it's done this way, because it's in a sample that people will reference and copy from. @shaoboyan could you help with this?

kainino0x (approved):

src/sample/videoUploadingWebCodecs/main.ts
shaoboyan: @kainino0x (OOO last Friday). Sry for late reply. Since we use pre-recorded video file as the input for VideoFrame, we need to use HTMLVideoElement to play it and call `captureStream` to get the VideoFrame. I considered constructing video frame in another way by using canvas and captureStream to canvas but it seems in that way we cannot cover multiple color space.

beaufortfrancois: I'm happy to remove the `getVideoFrameFromVideoElement` function and simply use `const videoFrame = new VideoFrame(video);` for this sample if it helps.

kainino0x: If it works for this sample I think we should.

@shaoboyan the color space issue is just due to the canvas, right? What about the `new VideoFrame(video)` codepath, does that work?

shaoboyan: >@shaoboyan the color space issue is just due to the canvas, right? 
Yes, due to canvas constraint color space.

> What about the new VideoFrame(video) codepath, does that work?
That would be ideal!

beaufortfrancois: Thank you! See https://github.com/webgpu/webgpu-samples/pull/251

PR #252, opened 2023-04-21
FIX updated cube uv value so texture won't be flipped

haxonadora:
Current uv values cause cube texture to be flipped. It's easier to notice when we use texture with text.
This PR updates uv value for `cubeVertexArray` so the texture will be displayed properly  

**Original texture:**
Di-3d
**Before**
Screenshot 2023-04-21 at 17 48 43
**After**
Screenshot 2023-04-21 at 18 06 00


kainino0x (approved):
Thanks!

PR #260, opened 2023-05-07
Added a Render Bundle sample

toji:
Saw someone asking about a sample of render bundles use last week and was kind of surprised to realize we didn't have one on here.  Coded up a fairly simple one, but I think it's enough to help developers understand the concept.

Screenshot 2023-05-07 at 2 08 19 PM

The page renders a planet with a ring of 5000 "asteroids" around it. For this demo each one is rendered with a separate draw call, which the demo points out is not the most efficient way to handle it but acts as a proxy for a larger, more complex scene so that the different in performance can be observed. There's a checkbox on the page to toggle between using the render bundle and rendering without one, and on my 2019 Macbook it makes **about a 10x difference** in the amount of JavaScript time spent each frame. (~0.7ms vs ~7ms).

In order to visualize the time spent per frame this sample adds [stats.js](https://github.com/mrdoob/stats.js/) as a dependency, as well as incorporating it into the sample framework in much the same way as dat.gui. It also adds a new utility file for generating sphere geometry which is largely a port of the [Three.js SphereGeometry logic](https://github.com/mrdoob/three.js/blob/master/src/geometries/SphereGeometry.js).

One bug that I can't figure out is that the checkmark in the dat.gui controls is getting doubled up, apparently because `init()` is getting called twice somehow? This doesn't seem to be a problem for any other samples, and I'm a bit stumped. I assume it's some react shenanigans that I'm not familiar with?


toji:
Okay, put a hack in place to prevent the dat.GUI controls from doubling up (though I still don't know why that's happening.) Also added a control for changing the number of asteroids in the scene (up to 10,000) , which can help emphasize the effect of the render bundle on faster devices. PTAL!

toji:
This PR is ready for review.

austinEng:
Sorry @toji, I haven't been able to look at this yet.

austinEng (approved):

PR #261, opened 2023-05-10
cornell: Fix 'raytracer' dispatch

ben-clayton:
This was rounding down the number of workgroups instead of 'ceil', resulting in a border at the right / bottom with garbage data.


ben-clayton:
Looks like GitHub actions got stuck.

PR #263, opened 2023-05-19
Fix TODO about gui duplication in renderBundles sample

kainino0x:


toji (approved):
Ah, so it was definitely a React-ism that I didn't understand. 😅

Thanks for the fix! LGTM.

PR #266, opened 2023-05-23
Black and White Filter

WUY97:


kainino0x (commented):
Thanks for the contribution! I have a number of comments on the code itself.

This sample is good on its own, but I'm not so sure about the added learning value to the samples. The unique part of this sample is just a few shader lines that implement the conversion to black-and-white. Yet, overall it's still a fair amount of code (which will need some maintenance), and I think all of the techniques here are already demonstrated by other samples. So it depends on other reviewers' opinions, but I would lean toward not including it - apologies if we decide not to.

WUY97:
Thank you for reviewing!! Your suggestions are very helpful for me to learn more about WebGPU. Really appreciate your help 😊

PR #267, opened 2023-05-23
Create ImageBitmaps from fetch instead of img.src

kainino0x:
createImageBitmap from fetch() is a best practice:

- HTMLImageElement isn't portable to Workers (helpful for anyone who wants to copy examples from here into their application).
- Creating an HTMLImageElement provides an incorrect signal to the browser that you want to put it in the DOM.
- img.decode() waits for the image to be ready to display in the DOM; it won't necessarily guarantee that there will be no synchronous blocking decode of the image in order to make the ImageBitmap.


shrekshao (approved):
Thanks for the explaination!

Issue #268, opened 2023-05-24
shadowMapping: wgpu-matrix 2.x changed mat4.lookAt()


microaaron:
https://github.com/webgpu/webgpu-samples/blob/0e1529da031b68e2ddf84a3a59651f2a4a987a11/src/sample/shadowMapping/main.ts#L367

https://github.com/webgpu/webgpu-samples/assets/16946786/e8455e05-4a92-4fa3-9c3b-98d4e4bce883

magcius:
In v1, wgpu-matrix's lookAt computes the inverse of what is commonly called lookAt in other matrix libraries. This was changed for v2. So this is correct for now, but would become a hazard when we upgrade.

https://github.com/greggman/wgpu-matrix/issues/2

microaaron:
> In v1, wgpu-matrix's lookAt computes the inverse of what is commonly called lookAt in other matrix libraries. This was changed for v2. So this is correct for now, but would become a hazard when we upgrade.
> 
> [greggman/wgpu-matrix#2](https://github.com/greggman/wgpu-matrix/issues/2)

Thanks. #271 is working on upgrading the samples to 2.x.

PR #269, opened 2023-05-24
shadowMapping: remove viewMatrix inversion (fixed by #271)

microaaron:
#268

https://github.com/webgpu/webgpu-samples/assets/16946786/4ee43d01-cc8a-41aa-b9a3-41fde87c9722

### DIFF
--- a/src/sample/shadowMapping/main.ts
+++ b/src/sample/shadowMapping/main.ts
@@ -292,9 +292,7 @@ const init: SampleInit = async ({ canvas, pageState }) => {
   const viewMatrix = mat4.inverse(mat4.lookAt(eyePosition, origin, upVector));
 
   const lightPosition = vec3.fromValues(50, 100, -100);
-  const lightViewMatrix = mat4.inverse(
-    mat4.lookAt(lightPosition, origin, upVector)
-  );
+  const lightViewMatrix = mat4.lookAt(lightPosition, origin, upVector);
 
   const lightProjectionMatrix = mat4.create();
   {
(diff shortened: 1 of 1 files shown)


kainino0x (changes requested):
It renders correctly before this patch, and incorrectly afterward. I can't tell offhand why these inverses are here, and I don't know if they can be removed. But if they can, more work is needed on this PR.

microaaron:
> It renders correctly before this patch, and incorrectly afterward. I can't tell offhand why these inverses are here, and I don't know if they can be removed. But if they can, more work is needed on this PR.

It seems to be alright after removing these inverses.

kainino0x:
I checked out your PR and found that it broke the rendering like so. Is your build updating correctly when you make code changes? It should do so automatically if you use `npm start`.
image

microaaron:
The build process seems to have made some changes to the source code, I'm not sure. I copied these source codes into a js file and ran it in the browser, and found this problem. But building according to the official method seems to be no problem.

microaaron:
Here is the example: [https://jsitor.com/r9ip6WtPnl](https://jsitor.com/r9ip6WtPnl). Key codes: line 408,411,483.

kainino0x:
The difference seems to be between wgpu-matrix 1.x and wgpu-matrix 2.x.

@greggman should we upgrade the samples to 2.x?

greggman:
If they're using wgpu-matrix then yes, and the breaking change from 1.x to 2.x is `mat4.lookAt` returns a view matrix (like other math libs).

microaaron:
Fixed by #271.

PR #270, opened 2023-05-25
deferredRendering: Switch from a positions buffer to reconstructing world-space positions from the depth buffer

magcius:
This saves on a buffer, and demonstrates how to reconstruct world-space positions from the depth buffer.

### DIFF
--- a/src/sample/deferredRendering/fragmentGBuffersDebugView.wgsl
+++ b/src/sample/deferredRendering/fragmentGBuffersDebugView.wgsl
@@ -18,7 +18,7 @@ fn main(
       vec2<i32>(floor(coord.xy)),
       0
     );
-    // remap depth into something a bit more palatable
+    // remap depth into something a bit more visible
     let depth = (1.0 - rawDepth) * 50.0;
     result = vec4(depth);
   } else if (c.x < 0.66667) {
--- a/src/sample/deferredRendering/fragmentGBuffersDebugView.wgsl
+++ b/src/sample/deferredRendering/fragmentGBuffersDebugView.wgsl
@@ -1,6 +1,7 @@
-@group(0) @binding(0) var gBufferPosition: texture_2d<f32>;
-@group(0) @binding(1) var gBufferNormal: texture_2d<f32>;
-@group(0) @binding(2) var gBufferAlbedo: texture_2d<f32>;
+
+@group(0) @binding(0) var gBufferNormal: texture_2d<f32>;
+@group(0) @binding(1) var gBufferAlbedo: texture_2d<f32>;
+@group(0) @binding(2) var gBufferDepth: texture_depth_2d;
 
 override canvasSizeWidth: f32;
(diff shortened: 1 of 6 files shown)


src/sample/deferredRendering/main.ts
```diff
@@ -553,6 +539,14 @@ const init: SampleInit = async ({ canvas, pageState, gui }) => {
...
       cameraViewProj.byteLength
     );
+    const cameraInvViewProj = mat4.invert(cameraViewProj) as Float32Array;
+    device.queue.writeBuffer(
+      cameraUniformBuffer,
+      64,
+      cameraInvViewProj.buffer,
+      cameraInvViewProj.byteOffset,
+      cameraInvViewProj.byteLength
+    );
```
toji, line 549: A comment more for @kainino0x and @shrekshao than @magcius: Seeing two `writeBuffers` to the same buffer like this was surprising to me, but looking through the other samples (and even this one) it turns out that's a semi-common pattern that we use.

I can't imagine that dispatching two queue operations is less overhead than copying the data into a larger `Float32Array` and doing a single write. As such, I feel like we should consider updating this and other similar snippets in the samples to do a single `writeBuffer` call when possible to help demonstrate best practices.

kainino0x: Ah, if only we could coalesce writes. :)
Yes, I think we should change these (can be after this PR). Theoretically these might be better served by buffer mapping, but it would be more complex and add allocation of GPUBuffers. I wish we had that old proposal I had that let you create ArrayBuffers in IPC-shared memory.

src/sample/deferredRendering/fragmentGBuffersDebugView.wgsl
```diff
@@ -12,11 +13,14 @@ fn main(
...
   let c = coord.xy / vec2<f32>(canvasSizeWidth, canvasSizeHeight);
   if (c.x < 0.33333) {
-    result = textureLoad(
-      gBufferPosition,
+    let rawDepth = textureLoad(
+      gBufferDepth,
       vec2<i32>(floor(coord.xy)),
       0
     );
+    // remap depth into something a bit more palatable
```
shrekshao, line 21: This seems to be for better visual purpose in debug view. I think it's better to mention the depth in debug view in sample description is flipped and remapped to clear confusion.

shrekshao (approved):
LGTM. I think the main benefit of using depth texture instead of a position buffer is to save g buffer bandwidth. As we already use float16 for normal to avoid hitting the maxColorAttachmentBytesPerSample limit, this does seem very useful.

src/sample/deferredRendering/fragmentGBuffersDebugView.wgsl
magcius: Changed.

toji:
Apologies! I thought this got merged a while ago.

PR #271, opened 2023-05-25
Use wgpu-matrix 2.x

### LINKED ISSUE #268: shadowMapping: wgpu-matrix 2.x changed mat4.lookAt()

greggman:
The breaking change was in 1.x, `mat4.lookAt` returned a camera matrix which was unexpected by most devs. 2.x changed `mat4.lookAt` to return a view matrix so that devs would get what they expect coming from other math libraries.

Closes #269
Fixes #268

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,7 +10,6 @@
       "license": "BSD-3-Clause",
       "dependencies": {
         "@types/dom-mediacapture-transform": "^0.1.5",
-        "@types/stats.js": "^0.17.0",
         "codemirror": "^5.58.2",
         "dat.gui": "^0.7.6",
         "file-loader": "^6.2.0",
@@ -19,7 +18,7 @@
         "react-dom": "^18.2.0",
         "stanford-dragon": "^1.1.1",
         "stats-js": "^1.0.1",
-        "wgpu-matrix": "^1.1.0"
+        "wgpu-matrix": "^2.4.0"
       },
       "devDependencies": {
         "@next/eslint-plugin-next": "^13.0.0",
@@ -28,6 +27,7 @@
         "@types/node": "^18.11.8",
         "@types/react": "^18.0.24",
         "@types/react-dom": "^18.0.8",
+        "@types/stats.js": "^0.17.0",
         "@typescript-eslint/eslint-plugin": "^5.41.0",
         "@typescript-eslint/parser": "^5.41.0",
         "@webgpu/types": "^0.1.21",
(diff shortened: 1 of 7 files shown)


austinEng (approved):

kainino0x (approved):

kainino0x:
LGTM but build failure.

greggman:
Sigh... type safety FTL (😂)

So pushed a fix but not sure I'm happy with it. Maybe it's rare though.

Here's the issue. The functions in the wgpu-matrix return `number[] | Float32Array | Float64Array` so if you do this

```
functionThatTakesAFloat32Array(mat4.identity());
```

You'll get an error since the types might not match.

I tried various solutions and didn't find one.

1st try was remove `number[]` from the types. That broke because you want to be able to pass native arrays. Example `vec3.normalize([1, 2, 3])`

2nd try was to change function parameters to `XXXLike`, example, `vec3.normalize(v: Vec3Like, dst?: Vec3): Vec3`

This allows you to pass in a native array, at least as the first argument. But, one sample is expecting to be able to pass a native array as `dst`

3rd try, change `dst` to also be `Vec3Like` as in vec3.normalize(v: Vec3Like, dst?: Vec3Like): Vec3. This requires casting inside the library functions from `return dst` to `return dst as Vec3` which is kind of a lie. You could pass a native array as `dst` and get one back but the return type says you're not.

It didn't actually solve the casting issue though since

```js
functionThatTakesAFloat32Array(mat4.identity());
```

Still expects only an `Float32Array` and was complaining it might be a `Float64Array`

4th try, put the wgpu-matrix code back the way it was and cast in the samples.

On the one had, it's kind of gross. On the the other hand, it's only coming up because of a uniform that has a single matrix. For most cases a uniform will have more than a single matrix in which case you'd use typed array views into the an arraybuffer and pass those as `dst` to the library and the issue goes away.  Or, you'd output from the library and then use `someUniformStructValueArray.set(someVec3, offset)` which should also work

So, for now it seems like the cast is best?

If you have any better ideas I'm happy to explore them.

kainino0x:
Could maybe parameterize the entire library so you can do something like:

```js
import wgpuMatrix from 'wgpu-matrix';
const { mat4, vec3 } = wgpuMatrix(Float32Array);
```

Just an off-the-cuff idea, might end up being very annoying. But aside from just the types, something like this would get rid of global statefulness of the library itself which I think is good because it will prevent two pieces of code in the same page from stepping each other's toes.

PR #273, opened 2023-05-30
a-buffer sample

rconde01:
Here's an A-Buffer sample. Looking forward to any feedback.

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,6 +25,7 @@
     "react-dom": "^18.2.0",
     "stanford-dragon": "^1.1.1",
     "stats-js": "^1.0.1",
+    "teapot": "^1.0.0",
     "wgpu-matrix": "^1.1.0"
   },
   "devDependencies": {
--- a/src/sample/a-buffer/composite.wgsl
+++ b/src/sample/a-buffer/composite.wgsl
@@ -1,0 +1,88 @@
+struct Uniforms {
+  modelViewProjectionMatrix: mat4x4<f32>,
+  maxStorableFragments: u32,
+  targetWidth: u32,
+};
+
+struct Heads {
+  numFragments: u32,
+  data: array<u32>
+};
+
+struct LinkedListElement {
+  color: vec4<f32>,
+  depth: f32,
+  next: u32
+};
+
+struct LinkedList {
+  data: array<LinkedListElement>
+};
+
+@binding(0) @group(0) var<uniform> uniforms: Uniforms;
+@binding(1) @group(0) var<storage,read_write> heads: Heads;
+@binding(2) @group(0) var<storage,read_write> linkedList: LinkedList;
+
+// Output a full screen quad
+@vertex
+fn main_vs(@builtin(vertex_index) vertIndex: u32) -> @builtin(position) vec4<f32> {
+  const position = array<vec2<f32>, 6>(
+    vec2(-1.0, -1.0),
+    vec2(1.0, -1.0),
+    vec2(1.0, 1.0),
+    vec2(-1.0, -1.0),
+    vec2(1.0, 1.0),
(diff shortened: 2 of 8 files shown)


package.json
```diff
@@ -25,6 +25,7 @@
     "react-dom": "^18.2.0",
     "stanford-dragon": "^1.1.1",
     "stats-js": "^1.0.1",
+    "teapot": "^1.0.0",
```
kainino0x, line 28: Haven't reviewed the rest yet, but just a thought. Let's pin these mesh dependencies at exact versions. I don't think we are ever going to want to update them (or deal with breakage if they have any), and it feels a little safer from an auditing perspective.

rconde01: Do you want a separate PR to pin the dragon...or just do both here?

kainino0x: Both here is fine, it's a small change.

src/sample/a-buffer/main.ts
```diff
@@ -0,0 +1,489 @@
...
+
+    const viewProjMatrix = mat4.multiply(projectionMatrix, viewMatrix);
+    return viewProjMatrix as Float32Array;
+  }
+
+  function frame() {
+    // Sample is no longer the active page.
+    if (!pageState.active) return;
+
+    // update the unforms buffer
```
kainino0x, line 395:
```suggestion
    // update the uniform buffer
```

src/sample/a-buffer/composite.wgsl
```diff
@@ -0,0 +1,88 @@
...
+  next: u32
+};
+
+struct LinkedList {
+  data: array<LinkedListElement>
+};
+
+@binding(0) @group(0) var<uniform> uniforms: Uniforms;
+@binding(1) @group(0) var<storage,read_write> heads: Heads;
+@binding(2) @group(0) var<storage,read_write> linkedList: LinkedList;
```
kainino0x, line 24:
```suggestion
@binding(1) @group(0) var<storage, read_write> heads: Heads;
@binding(2) @group(0) var<storage, read_write> linkedList: LinkedList;
```

src/sample/a-buffer/composite.wgsl
```diff
@@ -0,0 +1,88 @@
...
+
+@fragment
+fn main_fs(@builtin(position) position: vec4<f32>) -> @location(0) vec4<f32> {
+  let fragCoords = vec2<i32>(position.xy);
+  let headsIndex = u32(fragCoords.y) * uniforms.targetWidth + u32(fragCoords.x);
+
+  // The maximum layers we can process for any pixel
+  const maxLayers = 24u;
+
+  var layers: array<LinkedListElement,maxLayers>;
```
kainino0x, line 49:
```suggestion
  var layers: array<LinkedListElement, maxLayers>;
```

src/sample/a-buffer/translucent.wgsl
```diff
@@ -0,0 +1,87 @@
...
+  next: u32
+};
+
+struct LinkedList {
+  data: array<LinkedListElement>
+};
+
+@binding(0) @group(0) var<uniform> uniforms: Uniforms;
+@binding(1) @group(0) var<storage,read_write> heads: Heads;
+@binding(2) @group(0) var<storage,read_write> linkedList: LinkedList;
```
kainino0x, line 25:
```suggestion
@binding(1) @group(0) var<storage, read_write> heads: Heads;
@binding(2) @group(0) var<storage, read_write> linkedList: LinkedList;
```

kainino0x (approved):
LGTM, I couldn't review this in much depth since I don't know the algorithm, but it's very nice - love OIT! A substantial chunk of WGSL code so I think that makes it useful as a sample on its own.

PR #278, opened 2023-06-17
Update samples to work with TypeScript strict mode (part 1)

nash1111:
Like this, I want to add types to all samples to enhance safety. If there are no issues, I would like to proceed with adding them to all.

### DIFF
--- a/src/components/SampleLayout.tsx
+++ b/src/components/SampleLayout.tsx
@@ -136,6 +136,9 @@ const SampleLayout: React.FunctionComponent<
     };
     try {
       const canvas = canvasRef.current;
+      if (!canvas) {
+        throw new Error('The canvas is not available');
+      }
       const p = props.init({
         canvas,
         pageState,
--- a/src/components/SampleLayout.tsx
+++ b/src/components/SampleLayout.tsx
@@ -193,7 +193,9 @@ const SampleLayout: React.FunctionComponent<
         <p>{props.description}</p>
         {error ? (
           <>
-            <p>Is WebGPU Enabled?</p>
+            <p>
+              Something went wrong. Do your browser and device support WebGPU?
+            </p>
             <p>{`${error}`}</p>
           </>
         ) : null}
@@ -256,3 +258,9 @@ export const makeSample: (
 ) => JSX.Element = (props) => {
   return <SampleLayout {...props} />;
 };
+
+export function assert(condition: unknown, msg?: string): asserts condition {
+  if (!condition) {
+    throw new Error(msg);
+  }
+}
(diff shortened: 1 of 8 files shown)


src/components/SampleLayout.tsx
```diff
@@ -136,6 +136,9 @@ const SampleLayout: React.FunctionComponent<
     };
     try {
       const canvas = canvasRef.current;
+      if (!canvas) {
```
kainino0x, line 139: An assert helper will help a lot with cases like this and make the new checks less invasive.

You can use
```
export function assert(condition: unknown, msg?: string): asserts condition {
  if (!condition) {
    throw new Error(msg);
  }
}
```

Probably easiest to just put this in SampleLayout.tsx.

src/sample/animometer/main.ts
```diff
@@ -272,13 +282,13 @@ const init: SampleInit = async ({ canvas, pageState, gui }) => {
       }
     }
 
-    let startTime = undefined;
+    let startTime: number | undefined = undefined;
     const uniformTime = new Float32Array([0]);
 
     const renderPassDescriptor: GPURenderPassDescriptor = {
```
kainino0x: Committed suggestion to improve the typescript overhead here slightly.

src/sample/animometer/main.ts
```diff
@@ -4,6 +4,12 @@ import animometerWGSL from './animometer.wgsl';
 
 const init: SampleInit = async ({ canvas, pageState, gui }) => {
   const adapter = await navigator.gpu.requestAdapter();
+  if (!adapter) {
+    console.error(
+      'WebGPU is not supported. Make sure you are running the latest version of a compatible browser (like Chrome Canary) with the correct flags enabled.'
+    );
+    return;
```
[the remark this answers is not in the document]
kainino0x: (Done in the commit I pushed)

src/components/SampleLayout.tsx
kainino0x: (Done in the commit I pushed)

kainino0x:
Looking good so far, hopefully the suggestions will make updating the rest of the samples a bit easier - ping me when you need another review!

kainino0x:
And do be sure to pull the branch before you continue, because I committed suggestions to it.

nash1111:
Thank you for the review! I just did a git pull for your changes. Once I've corrected all the examples, I'll request a re-review from you!

nash1111:
I am capable of handling it, but given the volume, I'm considering splitting it into two PRs. The first PR will modify only the files that have been completed so far with strict: false, and the second will edit the remaining files to make the entire project strict: true. I believe that smaller PRs are preferable, so I'll be aiming for that. This PR is the first one. 
I will submit the second PR by the end of this month.

kainino0x (approved):
Sounds good! LGTM

PR #285, opened 2023-07-29
chore: wip adding type

nash1111:
im active. this PR will be ready for review soon

### DIFF
--- a/src/sample/cornell/main.ts
+++ b/src/sample/cornell/main.ts
@@ -11,6 +11,7 @@ import Radiosity from './radiosity';
 import Rasterizer from './rasterizer';
 import Tonemapper from './tonemapper';
 import Raytracer from './raytracer';
+import { assert } from '../../components/SampleLayout';
 
 const init: SampleInit = async ({ canvas, pageState, gui }) => {
   const presentationFormat = navigator.gpu.getPreferredCanvasFormat();
@@ -18,13 +19,14 @@ const init: SampleInit = async ({ canvas, pageState, gui }) => {
     presentationFormat === 'bgra8unorm' ? ['bgra8unorm-storage'] : [];
   const adapter = await navigator.gpu.requestAdapter();
   for (const feature of requiredFeatures) {
-    if (!adapter.features.has(feature)) {
+    if (!adapter?.features.has(feature)) {
       throw new Error(
         `sample requires ${feature}, but is not supported by the adapter`
       );
     }
   }
-  const device = await adapter.requestDevice({ requiredFeatures });
+  const device = await adapter?.requestDevice({ requiredFeatures });
+  assert(device, 'device is null');
 
   if (!pageState.active) return;
 
@@ -36,8 +38,8 @@ const init: SampleInit = async ({ canvas, pageState, gui }) => {
     rotateCamera: true,
   };
 
-  gui.add(params, 'renderer', ['rasterizer', 'raytracer']);
-  gui.add(params, 'rotateCamera', true);
+  gui?.add(params, 'renderer', ['rasterizer', 'raytracer']);
+  gui?.add(params, 'rotateCamera', true);
 
   const devicePixelRatio = window.devicePixelRatio || 1;
   canvas.width = canvas.clientWidth * devicePixelRatio;
@@ -80,7 +82,9 @@ const init: SampleInit = async ({ canvas, pageState, gui }) => {
     }
 
     const canvasTexture = context.getCurrentTexture();
-    const commandEncoder = device.createCommandEncoder();
+    const commandEncoder = device?.createCommandEncoder();
+    assert(commandEncoder, 'commandEncoder is null');
+    assert(device, 'device is null');
 
     common.update({
       rotateCamera: params.rotateCamera,
--- a/src/sample/cubemap/main.ts
+++ b/src/sample/cubemap/main.ts
@@ -1,5 +1,5 @@
 import { mat4, vec3 } from 'wgpu-matrix';
-import { makeSample, SampleInit } from '../../components/SampleLayout';
+import { assert, makeSample, SampleInit } from '../../components/SampleLayout';
 
 import {
   cubeVertexArray,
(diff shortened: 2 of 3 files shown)


src/sample/cubemap/main.ts
```diff
@@ -198,7 +199,7 @@ const init: SampleInit = async ({ canvas, pageState }) => {
   const renderPassDescriptor: GPURenderPassDescriptor = {
     colorAttachments: [
       {
-        view: undefined, // Assigned later
+        view: undefined as any, // Assigned later
```
kainino0x, line 202: (In response to the GitHub ACtions check warning)
It would be nice to make the build warning-free. In this case it can probably be `view: undefined!,`

src/sample/cornell/main.ts
```diff
@@ -11,20 +11,22 @@ import Radiosity from './radiosity';
...
+import { assert } from '../../components/SampleLayout';
 
 const init: SampleInit = async ({ canvas, pageState, gui }) => {
   const presentationFormat = navigator.gpu.getPreferredCanvasFormat();
   const requiredFeatures: GPUFeatureName[] =
     presentationFormat === 'bgra8unorm' ? ['bgra8unorm-storage'] : [];
   const adapter = await navigator.gpu.requestAdapter();
   for (const feature of requiredFeatures) {
-    if (!adapter.features.has(feature)) {
+    if (!adapter?.features.has(feature)) {
```
kainino0x, line 22: I think it would be better to `assert` the adapter first. Otherwise the error that's hit will be "sample requires feature" or "device is null" when in fact the adapter was null.

src/sample/cornell/main.ts
```diff
@@ -11,20 +11,22 @@ import Radiosity from './radiosity';
...
       throw new Error(
         `sample requires ${feature}, but is not supported by the adapter`
       );
     }
   }
-  const device = await adapter.requestDevice({ requiredFeatures });
+  const device = await adapter?.requestDevice({ requiredFeatures });
+  assert(device, 'device is null');
```
kainino0x, line 29: requestDevice itself can't actually return null, so if the `assert(adapter)` is added above then this assert won't be needed.

src/sample/cornell/main.ts
```diff
@@ -36,8 +38,8 @@ const init: SampleInit = async ({ canvas, pageState, gui }) => {
     rotateCamera: true,
   };
 
-  gui.add(params, 'renderer', ['rasterizer', 'raytracer']);
-  gui.add(params, 'rotateCamera', true);
+  gui?.add(params, 'renderer', ['rasterizer', 'raytracer']);
+  gui?.add(params, 'rotateCamera', true);
```
kainino0x, line 42: Since this can't actually happen I think I'd prefer `gui!` over `gui?`
If it's wrong it'll just throw a TypeError which is fine for a sample.

Or just `assert(gui)`.

src/sample/cornell/main.ts
```diff
@@ -80,7 +82,9 @@ const init: SampleInit = async ({ canvas, pageState, gui }) => {
     }
 
     const canvasTexture = context.getCurrentTexture();
-    const commandEncoder = device.createCommandEncoder();
+    const commandEncoder = device?.createCommandEncoder();
+    assert(commandEncoder, 'commandEncoder is null');
+    assert(device, 'device is null');
```
kainino0x, line 87: Neither of these can happen unless `device` is null. There's already an assert above that `device` isn't null, so the `?` and both of these asserts shouldn't be necessary.

src/sample/cubemap/main.ts
```diff
@@ -245,6 +246,7 @@ const init: SampleInit = async ({ canvas, pageState }) => {
   function frame() {
     // Sample is no longer the active page.
     if (!pageState.active) return;
+    assert(device, 'device is null');
```
kainino0x, line 249: Similarly shouldn't be necessary, though I'm less certain about this since it's inside a function. OK to keep if needed.

src/sample/cubemap/main.ts
```diff
@@ -255,9 +257,16 @@ const init: SampleInit = async ({ canvas, pageState }) => {
...
+    type GPURenderPassColorAttachmentArray =
+      (GPURenderPassColorAttachment | null)[];
+
+    const attachment = (
+      renderPassDescriptor.colorAttachments as GPURenderPassColorAttachmentArray
+    )[0];
+
+    assert(attachment, 'attachment is null');
+
+    attachment.view = context.getCurrentTexture().createView();
```
kainino0x, line 269: This is a complicated way to get `colorAttachments[0]`. How about storing colorAttachment0 into its own variable?

```ts
const colorAttachment0: GPURenderPassColorAttachment = {
  view: undefined as any, // Assigned later
};

const renderPassDescriptor: GPURenderPassDescriptor = {
  colorAttachments: [colorAttachment0],
```
```ts
colorAttachment0.view = context.getCurrentTexture().createView();
```

kainino0x (commented):
Overall this is adding too much lines-of-code overhead to make TypeScript happy, which isn't good for readability. Some overhead is needed, but I think we can reduce it.

I've commented on a few things that should significantly improve the readability. Most of my suggestions apply to many places later in the PR - please apply them in all similar situations. (Or update one or two and ping me, if you'd like an incremental review of particular changes before you go through and do all of them.)

nash1111:
Thank you for the review & I apologize for the late response. I will incorporate your feedback.

kainino0x:
Thanks, no rush, thank you for the contribution!

src/sample/cubemap/main.ts
nash1111: I wanted to use ```view: undefined!``` too, but it was inferred as the never type and didn't work out. How about using ```view: undefined as unknown as GPUTextureView``` ? Using as twice isn't intuitive, but since we can't directly cast undefined to GPUTextureView, it seems unavoidable. If there's a better suggestion, I'd love to hear it.

PR #289, opened 2023-08-04
Added a sample that shows WebGPU running from a web worker

toji:
This is a pretty simple sample, but it's covering a subject that I've seen a couple of questions about and I think i's useful to have readily available as a sanity check to ensure this method of using WebGPU works.

The sample itself is basically just a copy of the rotating cube sample, but with all the WebGPU code executing from a worker rather than the main thread. `canvas.transferToOffscreen()` is used to produce an `OffscreenCanvas` on the main thread that's then transferred to the worker and used for rendering. The frame loop is managed worker-side with `requestAnimationFrame()`.

If we feel it's useful we could set up variants/options for this sample that use the `transferToImageBitmap()` pattern for rendering and a main-thread-driven rendering loop, but I think it's probably OK to have it focus on simply demonstrating that you can use WebGPU in a worker.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,7 +10,6 @@
       "license": "BSD-3-Clause",
       "dependencies": {
         "@types/dom-mediacapture-transform": "^0.1.5",
-        "@types/stats.js": "^0.17.0",
         "codemirror": "^5.58.2",
         "dat.gui": "^0.7.6",
         "file-loader": "^6.2.0",
@@ -28,6 +27,7 @@
         "@types/node": "^18.11.8",
         "@types/react": "^18.0.24",
         "@types/react-dom": "^18.0.8",
+        "@types/stats.js": "^0.17.0",
         "@typescript-eslint/eslint-plugin": "^5.41.0",
         "@typescript-eslint/parser": "^5.41.0",
         "@webgpu/types": "^0.1.21",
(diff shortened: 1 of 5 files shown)


toji:
Hm... build checks think that `transferControlToOffscreen()` doesn't exist. [CanIUse begs to differ.](https://caniuse.com/?search=transferControlToOffscreen) 😉

Any browser that implements WebGPU will also have this API, so I think it's a non-issue. I'll look into how to suppress that error later, probably Monday.

kainino0x:
> Hm... build checks think that `transferControlToOffscreen()` doesn't exist. [CanIUse begs to differ.](https://caniuse.com/?search=transferControlToOffscreen) 😉

It was removed in TypeScript 4.4
https://github.com/microsoft/TypeScript/commit/9ebe11c2d375bfb1c8fad10d705ffae09583c522
https://github.com/microsoft/TypeScript-DOM-lib-generator/issues/1029
and added back in 4.9.
https://github.com/microsoft/TypeScript/commit/170a17fad57eae619c5ef2b7bdb3ac00d6c32c47

Should just need to upgrade to 4.9 or newer.

toji:
Done. updated to the last 4.x revision (4.9.5)

toji:
Friendly Ping!

austinEng (approved):

PR #297, opened 2023-09-14
Clamp devicePixelRatio in a-buffer sample

### LINKED ISSUE #294: A-Buffer sample fails on some Android devices
On both a Pixel 4 and Pixel 6 (but not a Pixel 7) the A buffer sample fails with the following error:

```
Binding size (171606624) of [Buffer "linkedListBuffer"] is larger than the maximum binding size (134217728).
 - While validating entries[2] as a Buffer.
Expected entry layout: { binding: 2, visibility: ShaderStage::Fragment, buffer: { type: BufferBindingType::Storage, hasDynamicOffset: 0, minBindingSize: 32 } }
 - While validating [BindGroupDescriptor "translucentBindGroup"] against [BindGroupLayout]
 - While calling [Device].CreateBindGroup([BindGroupDescriptor "translucentBindGroup"]).
```

Given that the [size of the `linkedListBuffer`](https://github.com/webgpu/webgpu-samples/blob/main/src/sample/a-buffer/main.ts#L111) only differs between devices based on the canvas size, I went searching as was surprised to find that the Pixel 4 and 6 had a `devicePixelRatio` of 3.5, but the Pixel 7 I was testing only had a  `devicePixelRatio` of 2.5.

The buffer size/canvas resolution for this sample should be capped to stay under the binding limit.

toji:
Fixes #294

Prevents the sample from failing on devices where a high `devicePixelRatio` (Like 3.5) would cause the `linkedListBuffer` from exceeding  `maxStorageBufferBindingSize` (128Mib).

### DIFF
--- a/src/sample/a-buffer/main.ts
+++ b/src/sample/a-buffer/main.ts
@@ -18,7 +18,11 @@ const init: SampleInit = async ({ canvas, pageState }) => {
   if (!pageState.active) return;
   const context = canvas.getContext('webgpu') as GPUCanvasContext;
 
-  const devicePixelRatio = window.devicePixelRatio || 1;
+  // The devicePixelRatio is clamped here because the linkedListBuffer size will
+  // be computed based on the canvas size, but if the devicePixelRatio is too
+  // high then it will cause that buffer to exceed the default
+  // maxStorageBufferBindingSize (128Mib).
+  const devicePixelRatio = Math.min(window.devicePixelRatio || 1, 3);
   canvas.width = canvas.clientWidth * devicePixelRatio;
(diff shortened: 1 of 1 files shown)


toji:
Review ping?

austinEng (approved):
LGTM if you want to land it. but it seems like rconde will have an alternative fix that doesn't decreases the resolution

https://github.com/webgpu/webgpu-samples/issues/294#issuecomment-1732708050

toji:
Landing for now as a short term fix, and possibly that could serve as an alternate code path to @rconde01's suggestion? I would consider it useful to help devs understand that both are valid ways to work around the limit with obvious tradeoffs.

PR #300, opened 2023-09-28
Enhance A-buffer sample with ability to split into multiple passes to limit memory usage

rconde01:
This is based on the discussion [here](https://github.com/webgpu/webgpu-samples/issues/294#issuecomment-1736252524). This also allows the user to select either multipass or devicePixelRatio clamping as the limiting strategy.

### DIFF
--- a/src/sample/a-buffer/main.ts
+++ b/src/sample/a-buffer/main.ts
@@ -75,18 +75,77 @@ const init: SampleInit = async ({ canvas, pageState }) => {
     indexBuffer.unmap();
   }
 
+  // Determines how much memory is allocated to store linked-list elements
+  const averageLayersPerFragment = 4;
+
+  // Each element stores
+  // * color : vec4<f32>
+  // * depth : f32
+  // * index of next element in the list : u32
+  const linkedListElementSize =
+    5 * Float32Array.BYTES_PER_ELEMENT + 1 * Uint32Array.BYTES_PER_ELEMENT;
+
+  let numSlices = 1;
+  let sliceHeight = canvas.height;
+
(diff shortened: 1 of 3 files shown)


src/sample/a-buffer/main.ts
```diff
@@ -340,122 +331,345 @@ const init: SampleInit = async ({ canvas, pageState }) => {
...
+    const calculateLinkedListBufferSize = (width: number, height: number) => {
+      return averageLayersPerFragment * linkedListElementSize * width * height;
+    };
+
+    let linkedListBufferSize = calculateLinkedListBufferSize(
+      canvas.width,
+      canvas.height
+    );
+
+    while (linkedListBufferSize > device.limits.maxStorageBufferBindingSize) {
```
greggman, line 399: Could this be something along the lines of

```
  lineSize = averageLayersPerFragment * linkedListElementSize * width
  maxLines = Math.floor(device.limits.maxStorageBufferBindingSize / lineSize)
  numSlices = Math.ceil(height / maxLines)
  sliceHeight = Math.ceil(canvas.height / numSlices);
```

or something similar? It's not clear to me why this calculation needs to be iterative.

greggman (commented):
FYI: I can get a dpr of 5 just by pressing Cmd/Ctrl + (zoom in) to the limit on my desktop. No need for mobile for this to be an issue

greggman (approved):
lgtm

src/sample/a-buffer/main.ts
rconde01: Comments should be addressed...let me know if there's anything else

PR #305, opened 2023-10-18
Added normal mapping example

cmhhelgeson:
An example of normal mapping that allows the user to adjust the lighting of the scene, the depth layers, and the currently displayed combination of diffuse, normal, and height maps (Spiral, Toybox, Bark).

Current issues to be addressed:
Parallax Occlusion Mapping (Possibly in a later PR)
Weird Bark Height Map Behavior
Creating Generalized Way of Loading Images in Assets Folder from Helper Functions Rather Than Public
Removing unused types, interfaces, and helper functions per usual.


cmhhelgeson:
Ready for review. Contains some of the same utils as used in the bitonicSort shader, which can either be inlined, kept, or moved to a separate folder depending on the maintainers' preferences.

cmhhelgeson:
Adding video to demonstrate intended functionality.

https://github.com/webgpu/webgpu-samples/assets/62450112/116be3a4-b037-46ac-84a9-aac80b4edd0e

austinEng:
@greggman could you help review this PR?

src/sample/normalMap/normalMap.wgsl
```diff
@@ -0,0 +1,207 @@
...
+  @location(3) posTS: vec3f,
+  // View position in tangent space
+  @location(4) viewTS: vec3f,
+  // Extracted components of our tbn matrix
+  @location(5) tbnTS0: vec3<f32>, 
+  @location(6) tbnTS1: vec3<f32>,
+  @location(7) tbnTS2: vec3<f32>,
+}
+
+fn transpose3x3(mat: mat3x3f) -> mat3x3f  {
```
greggman, line 42: `transpose` is a built in WGSL function.

src/sample/normalMap/normalMap.wgsl
```diff
@@ -0,0 +1,207 @@
...
+@group(1) @binding(2) var normalTexture: texture_2d<f32>;
+@group(1) @binding(3) var depthTexture: texture_2d<f32>;
+
+fn parallax_uv(
+  uv: vec2f, 
+  viewDirTS: vec3f, 
+  depthSample: f32,
+  depthScale: f32,
+) -> vec2f {
+  if (mapInfo.mappingType == 4) {
```
greggman, line 66: Are uber shaders generally considered bad practice? Would this be better with a separate shader for each mode? I'd think we don't want people copying this example if it's not considered a good thing to do. But I might wrong.

src/sample/normalMap/normalMap.wgsl
```diff
@@ -0,0 +1,207 @@
...
+    // Perturb uv coordinates based on depth and camera direction
+    var p: vec2f = viewDirTS.xy * (depthSample * depthScale) / viewDirTS.z;
+    return uv - p;
+  }
+  // Break up depth space into layers
+  var depthPerLayer: f32 = 1.0 / f32(mapInfo.depthLayers);
```
greggman, line 72: Should you use `let` for things that never change, instead of `var`?

src/sample/normalMap/normalMap.wgsl
```diff
@@ -0,0 +1,207 @@
...
+  // Start at lowest depth
+  var currentDepth: f32 = 0.0;
+  var delta_uv: vec2<f32> = viewDirTS.xy * depthScale / (viewDirTS.z * mapInfo.depthLayers);
```
greggman, line 75: you don't generally need to specify types on variables as they inherit their type from the type on the right. Here and a bunch of other places would it be better to just infer the types? It seems inconsistent where some places they are inferred and others not?

src/sample/normalMap/utils.ts
```diff
@@ -0,0 +1,264 @@
...
+  visibilities: number[],
+  resourceTypes: ResourceTypeName[],
+  resourceLayouts: BindGroupBindingLayout[],
+  resources: GPUBindingResource[][],
+  label: string,
+  device: GPUDevice
+): BindGroupsObjectsAndLayout => {
+  const layoutEntries: GPUBindGroupLayoutEntry[] = [];
+  for (let i = 0; i < bindings.length; i++) {
+    const layoutEntry: any = {};
```
greggman, line 37: This section can be changed to this to remove the `any`

```js
  for (let i = 0; i < bindings.length; i++) {
    layoutEntries.push({
      binding: bindings[i],
      visibility: visibilities[i % visibilities.length],
      [resourceTypes[i]]: resourceLayouts[i],
    });
  }
```

src/sample/normalMap/utils.ts
```diff
@@ -0,0 +1,264 @@
...
+  const bindGroupLayout = device.createBindGroupLayout({
+    label: `${label}.bindGroupLayout`,
+    entries: layoutEntries,
+  });
+
+  const bindGroups: GPUBindGroup[] = [];
+  for (let i = 0; i < resources.length; i++) {
+    const groupEntries: GPUBindGroupEntry[] = [];
+    for (let j = 0; j < resources[0].length; j++) {
+      const groupEntry: any = {};
```
greggman, line 53: same as above

src/meshes/box.ts
```diff
@@ -0,0 +1,347 @@
...
+    depthSegments
+  );
+
+  const originalStrideElements =
+    mesh.vertexStride / Float32Array.BYTES_PER_ELEMENT;
+
+  const vertexCount = mesh.vertices.length / originalStrideElements;
+
+  const tangents = new Array(vertexCount);
+  tangents.fill(vec3.create(0.0, 0.0, 0.0));
```
greggman, line 248: I'm not sure this is doing what you think it's doing. It's making a single `vec3` and putting a reference to it in every slot of the array.

In other words, then this is done, 

```
tangents[12][0] = 456;
console.log(tangents[47][0]); // prints 456 
```

Is that what you wanted?

Also, there's no good reason to call `vec3.create(0.0, 0.0, 0.0)`. You can just do `vec3.create()` or better just `[0, 0, 0]`. It will be much faster, here and other places.

src/sample/normalMap/main.ts
```diff
@@ -0,0 +1,407 @@
...
+  Toybox,
+  BrickWall,
+}
+
+const init: SampleInit = async ({ canvas, pageState, gui }) => {
+  const adapter = await navigator.gpu.requestAdapter();
+  const device = await adapter.requestDevice();
+  if (!pageState.active) return;
+  const context = canvas.getContext('webgpu') as GPUCanvasContext;
+  const devicePixelRatio = window.devicePixelRatio || 1;
```
greggman, line 25: there's no need for the `|| 1`.  That's left over from like 2011. No browser the supports WebGPU would not have `devicePixelRatio`

src/sample/normalMap/main.ts
```diff
@@ -0,0 +1,407 @@
...
+  const lightPosXCell = lightFolder.add(settings, 'lightPosX', -5, 5).step(0.1);
+  const lightPosYCell = lightFolder.add(settings, 'lightPosY', -5, 5).step(0.1);
+  const lightPosZCell = lightFolder.add(settings, 'lightPosZ', -5, 5).step(0.1);
+  const lightIntensityCell = lightFolder
+    .add(settings, 'lightIntensity', 0.0, 0.1)
+    .step(0.002);
+  depthFolder.add(settings, 'depthScale', 0.0, 0.1).step(0.01);
+  depthFolder.add(settings, 'depthLayers', 1, 32).step(1);
+
+  const liFunctionElements = document.getElementsByClassName('cr function');
```
greggman, line 297: Hacking deep into the GUI library's elements is brittle. This should probably be removed.

greggman (changes requested):
So I don't know how opinionated to be on this review. Please excuse me if I'm being too strict. I'm just going with my own thoughts. I'm also unfamiliar with the other samples. They may have similar issues.

I ran it locally and ran into the following issues

* The cube is too far away to see any details

To fix I set the camera to this

```js
    cameraPosX: 0.0,
    cameraPosY: 0.8,   // was 0
    cameraPosZ: -1.4,  // was -2.4
```

I removed the x rotation from the cube since just moving the camera above the cube seemed less surprising

```js
    // mat4.rotateX(modelMatrix, 10, modelMatrix);
```

and updated the perspective to 0.1 instead of 1 for near

```js
  const projectionMatrix = mat4.perspective(
    (2 * Math.PI) / 5,
    aspect,
    0.1,   // was 1
    10.0   // was 100
  ) as Float32Array;
```

* The view matrix seemed wrong. I changed to this

```js
  function getViewMatrix() {
    return mat4.lookAt(
      [settings.cameraPosX, settings.cameraPosY, settings.cameraPosZ],
      [0, 0, 0],
      [0, 1, 0]
    );
  }
```

* The light seemed wrong. 

If the cube is at 0,0,0 and the camera is -Z then a light at

```js
    lightPosX: 1.7,
    lightPosY: -0.7,
    lightPosZ: 1.9,
```

Would be behind the cube (camera is -Z, light is +Z), to right (X is positive), and under the cube (Y is negative).

Changed to

```js
    lightPosX: 1.7,
    lightPosY: 0.7,   // above the cube
    lightPosZ: -1.9,  // same side as camera
```

* Then, I couldn't see any difference in rendering between these "Bump Mode"s

    | 'Normal Texture'
    | 'Depth Texture'
    | 'Normal Map'
    | 'Parallax Scale'
    | 'Steep Parallax';

To make sure I wasn't crazy I added an 'animate' GUI setting that stops the rotation. To do that I added a clock

```js
  const settings: GUISettings = {
    ...
    animate: true,
    ...
  };

  ...

  gui.add(settings, 'animate');

  ...

  let then = 0;
  let time = 0;
  function frame(now) {
    if (!pageState.active) return;

    const elapsedTime = settings.animate ? (then - now) * 0.001 : 0;
    time += elapsedTime;
    then = now;

    // Write to normal map shader
    const viewMatrix = getViewMatrix();
    const modelMatrix = getModelMatrix(time);
```

and updated `getModelMatrix` to

```js

  function getModelMatrix(time: number) {
    const modelMatrix = mat4.create();
    mat4.identitåy(modelMatrix);
    mat4.rotateY(modelMatrix, time * -0.5, modelMatrix);
    return modelMatrix;
  }
```

Once I had that in I tried stopping the animation and switching the bump mode. There is no difference in rendering between any of them except for "Diffuse Texture"


https://github.com/webgpu/webgpu-samples/assets/234804/11de13f4-cc3a-4e83-b3f1-5167eb252285

* I'm not sure why all the casts like this

```js
    // Write to normal map shader
    const viewMatrixTemp = getViewMatrix();
    const viewMatrix = viewMatrixTemp as Float32Array;

    const modelMatrixTemp = getModelMatrix();
    const modelMatrix = modelMatrixTemp as Float32Array;
```

I changed to 

```js
    // Write to normal map shader
    const viewMatrix = getViewMatrix();
    const modelMatrix = getModelMatrix(time);
```

And I don't get any TypeScript warnings 🤷‍♂️

Here's the changes I made mentioned above

https://github.com/greggman/webgpu-samples/commit/db27f546eb7ddb2346e00276127bfe79075ba6d1

src/sample/normalMap/normalMap.wgsl
cmhhelgeson: Fixed

src/meshes/box.ts
```diff
@@ -0,0 +1,347 @@
...
+  );
+
+  const vertexStride = 8 * Float32Array.BYTES_PER_ELEMENT; //calculateVertexStride(vertexProperties);
+
+  const indicesArray =
+    indexFormat === 'uint16'
+      ? new Uint16Array(indices)
+      : new Uint32Array(indices);
+
+  console.log(indicesArray);
```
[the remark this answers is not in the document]
cmhhelgeson: Removed.

src/meshes/box.ts
cmhhelgeson: Fixed

src/meshes/box.ts
```diff
@@ -0,0 +1,347 @@
...
+    const deltaUV2 = vec3.sub(uv3, uv1);
+
+    // Edge of a triangle moves in both u and v direction (2d)
+    // deltaU * tangent vector + deltav * bitangent
+    // Manipulating the data into matrices, we get an equation
+
+    const constantVal =
+      1.0 / (deltaUV1[0] * deltaUV2[1] - deltaUV1[1] * deltaUV2[0]);
+
+    const tangent = vec3.fromValues(
```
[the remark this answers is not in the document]
cmhhelgeson: Fixed

src/components/SampleLayout.module.css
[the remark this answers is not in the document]
cmhhelgeson: Removed

src/sample/normalMap/utils.ts
cmhhelgeson: Fixed

src/sample/normalMap/utils.ts
cmhhelgeson: Fixed

src/sample/normalMap/normalMap.wgsl
cmhhelgeson: Made this consistent by removing type references when unecessary.

src/sample/normalMap/normalMap.wgsl
cmhhelgeson: I think it's more understandable and performant than having to manage multiple pipelines, especially when different types of depth mapping are still using the same code, but I defer to your expertise.

src/sample/normalMap/main.ts
cmhhelgeson: I'll remove it, but we'd also have to remove it from bitonicSort as well.

src/sample/normalMap/normalMap.wgsl
cmhhelgeson: Fixed

cmhhelgeson:
I implemented most of the suggestions (although the perspective of my camera doesn't seem to match yours). However, I left in the model rotation, since I think it more effectively demonstrates how the perceptual depth changes as the surface moves in and out of light, and I wasn't sure what was meant by surprising (too visually distracting, unexpected?).

cmhhelgeson:
Also, I was able to fix the issue with the map mode not changing by remembering to write that value as a u32 rather than a f32.

src/sample/normalMap/main.ts
greggman: I will fix the other samples. Thanks for pointing that out.

greggman (approved):
lgtm

PR #309, opened 2023-10-27
Bitonic Sort Update

cmhhelgeson:
An update to the bitonic sort example with a few new features: (UPDATED 10/30/2023)

1. Initial sort automatically executes after page load and the speed of the sort can be dynamically adjusted. 

2. Will be to use multiple workgroups to sort more elements than the max number of threads per workgroup will allow. The program will also display, based on the number of elements, how many workgroups are dispatched per step, and the number of workgroups will dynamically update depending on the number of elements.

3. Added an atomic counter that tracks the number of swaps that have been made during a sort. 

4. Various naming and GUI tweaks to decrease clutter and increase clarity including:
          a. GUI elements are now named controllers rather than cells to both match with Dat.GUI's types and to decrease confusion when referring to grid display cells.
          b. GUI elements that should be non-interactive have been made non-interactive
          c. GUI elements related to the changes above have been added

~Pull request is not yet ready to review~ (UPDATE 10/30/2023 Ready for review)

### DIFF
--- a/src/sample/bitonicSort/atomicToZero.wgsl
+++ b/src/sample/bitonicSort/atomicToZero.wgsl
@@ -1,0 +1,7 @@
+@group(0) @binding(3) var<storage, read_write> counter: atomic<u32>;
+
+@compute @workgroup_size(1, 1, 1)
+fn atomicToZero() {
+  let counterValue = atomicLoad(&counter);
+  atomicSub(&counter, counterValue);
+}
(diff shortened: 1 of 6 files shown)


greggman (approved):
lgtm

I'm not super enthralled with the hacking the GUI but I'll work on providing a better solution and let this go for now.

cmhhelgeson:
Note: Changes were approved 3 days ago, but additional changes have come in that have yet to be reviewed.

cmhhelgeson:
Made one more minor update that just removes interactivity from GUI elements and adds an atomic counter for the number of swaps made during a sort (ie the number of times idx.x > idx.y). 

Last thing I'm adding. Definitely.

PR #310, opened 2023-10-27
Fix image loading in normalMap sample

cmhhelgeson:
Get the image loading in the normalMap sample to actually work.


cmhhelgeson:
How are assets provided to the static build of the Next project? Adding images to assets and then running build or export doesn't seem to add them to either the static build or the out build.

cmhhelgeson:
Got rid of any utils outside of the main function that handled loading images. I also put all the images back in the assets folder. For some reason, the .next folder and out folder only build in the assets as part of the website when the images are referenced in main? Not sure why that might be. Regardless, image loading should be fixed after this next commit.

cmhhelgeson:
This should probably be addressed as soon as possible, the current example on the website is nonfunctional.

next.config.js
```diff
@@ -8,6 +8,9 @@ module.exports = {
   basePath: BASE_PATH,
   compress: true,
   reactStrictMode: true,
+  eslint: {
```
greggman, line 11: remove?

src/sample/normalMap/main.ts
```diff
@@ -219,7 +284,7 @@ const init: SampleInit = async ({ canvas, pageState, gui }) => {
   function getModelMatrix() {
     const modelMatrix = mat4.create();
     mat4.identity(modelMatrix);
-    mat4.rotateX(modelMatrix, 10, modelMatrix);
+    mat4.rotateX(modelMatrix, 0, modelMatrix);
```
greggman, line 287: why rotate by 0? remove?

src/sample/texturedCube/main.ts
```diff
@@ -110,6 +110,11 @@ const init: SampleInit = async ({ canvas, pageState }) => {
   // Fetch the image and upload it into a GPUTexture.
   let cubeTexture: GPUTexture;
   {
+    const url = new URL(
```
greggman, line 113: why is this changed? revert?

greggman (changes requested):

src/sample/normalMap/main.ts
cmhhelgeson: Fixed

next.config.js
cmhhelgeson: Removed

greggman (approved):

PR #311, opened 2023-10-28
Workers can log to the console just fine.

greggman:
Did you mean to say something else?


toji (approved):
I legitimately thought that you still couldn't log from a worker. It was definitely a restriction at some point, but I'm glad to see that it works at this point. Thank you for catching my use of an outdated workaround!

Issue #317, opened 2023-10-31
Normal mapping sample doesn't look quite right


ben-clayton:
The normals on the toybox look backwards.

Move the light so it is up high (`lightPosY=2.3`).
According to the parallax mode, the pyramid is supposed to stand outwards. The light is positioned above the box, yet the normals suggest the opposite side of the pyramid is more lit.

<img width="496" alt="Screenshot 2023-10-31 at 19 55 44" src="https://github.com/webgpu/webgpu-samples/assets/11505236/cc9a925a-ce3d-4d6f-9cda-85a3a4a52dd7">

ben-clayton:
@cmhhelgeson - FYI

cmhhelgeson:
Sorry for late response, will look into it.

ben-clayton:
I think you just need to mirror the normal map in the X and Y directions.

cmhhelgeson:
Taking a break today to deal with some sickness, but if it's a simple fix then feel free to change it, otherwise I can come back to it later.

PR #319, opened 2023-11-07
Flipped toybox normal map along x and y-axis.

cmhhelgeson:
See issue #317


ben-clayton:
It looks like after you flipped the height map, you then regenerated the normal-map, instead of just flipping that. For some reason, the normals for this texture then point the wrong way. I don't know why this texture appears wrong, when the others look okay.

I wanted to confirm that this was an issue with the texture and not a bug in the code, so I spent some time going through the sample's code. I think the code was sound, but I cleaned up the code as I went along. PR here: #320

cmhhelgeson:
Pull request closed due to issues being more comprehensively addressed by #320.

PR #320, opened 2023-11-07
Simplify normal map sample, fix toy box normals.

ben-clayton:
Mirror the toy-box texture to fix the normals.

Simplify the `box.ts` code.
This was more complicated than it needed to be, and borrowed code from another project.
This code also used `vec3` methods with two-element vectors, which might cause issues with future versions of `wgpu-matrix`

Simplfiy `normalMap.wgsl`
* Use `const` declarations to name the various modes instead of magic literals.
* Calculate the `worldViewProj` and `worldView` matrices on the CPU and pass them in as part of the uniform buffer. Reduces duplicated work in the vertex shader, and reduces the amount of data needed to be passed in.
* Use `vec3f` for the light position instead of scalar. Pass in the position in view-space.
* Reduce the amount of data passed from vertex to fragment.
* Rename `tbn` to `tangent`, `bitangent` and `normal`.
* Remove unused function `when_greater`
* Split `parallax_uv` into two functions.
* Simplify steep-parallax loop. Pre-calculate the UV derivatives so we can break early from the loop.

Rename `diffuse` textures to `albedo`. Technically a diffuse texture already has diffuse lighting applied.

Fixed: #317


ben-clayton:
FYI: @cmhhelgeson

Kangz (approved):

cmhhelgeson:
Hi Ben,

I took a look at the code and everything looks good in terms of clarity and simplicity. Sorry I couldn't be of more help, I've been sick for longer than expected.

ben-clayton:
> Hi Ben,
> 
> I took a look at the code and everything looks good in terms of clarity and simplicity. Sorry I couldn't be of more help, I've been sick for longer than expected.

No need to apologise! Thank you for being open to the changes.
I ended up looking through this code because identifying these kinds of bugs are always horribly tricky. My ex-gamedev eyes knew _something_ was off, but it was certainly hard to pin down.
Thank you for submitting the sample. It's great.

PR #322, opened 2023-11-09
Added Tab Animation to Code Source Hover/Data-Active=True

cmhhelgeson:
https://github.com/webgpu/webgpu-samples/assets/62450112/a1b3b817-e34d-4d63-af67-5fe2925887bc

The animation does use the CSS :has property, but all devices that support WebGPU should be able to access that property. Readability of sources also doesn't seem effected even when the window size is cramped.


cmhhelgeson:
https://github.com/webgpu/webgpu-samples/assets/62450112/a288a1e4-aaeb-46ca-bed2-104fcbbbe611

Adding more representative video of switching between tabs.

kainino0x (approved):
nice little bit of polish, LGTM!

PR #323, opened 2023-11-09
Add timestamp query support to compute boids

beaufortfrancois:
This PR adds timestamp query support to the compute boids sample by showing render pass and compute pass durations. 

I've tried several approaches before this attempt:
- Creating a queryset each frame but it fails very quickly on macOS  with [DAWN_OUT_OF_MEMORY_ERROR](https://source.chromium.org/chromium/chromium/src/+/main:third_party/dawn/src/dawn/native/metal/QuerySetMTL.mm;l=71;drc=e542c349d58a443b323ac97843739c8264030df3).
- Waiting for `resultBuffer.mapAsync` to settle in `frame()` but this makes UI janky as expected.

image

### DIFF
--- a/src/sample/computeBoids/main.ts
+++ b/src/sample/computeBoids/main.ts
@@ -120,20 +120,10 @@ const init: SampleInit = async ({ canvas, pageState, gui }) => {
   const computePassDescriptor = {};
 
   let querySet: GPUQuerySet | undefined = undefined;
-  let resolveBuffer: GPUBuffer | undefined = undefined;
-  let resultBuffer: GPUBuffer | undefined = undefined;
   if (hasTimestampQuery) {
     querySet = device.createQuerySet({
       type: 'timestamp',
       count: 4,
-    });
-    resolveBuffer = device.createBuffer({
-      size: 4 * BigInt64Array.BYTES_PER_ELEMENT,
-      usage: GPUBufferUsage.QUERY_RESOLVE | GPUBufferUsage.COPY_SRC,
-    });
-    resultBuffer = device.createBuffer({
-      size: 4 * BigInt64Array.BYTES_PER_ELEMENT,
-      usage: GPUBufferUsage.COPY_DST | GPUBufferUsage.MAP_READ,
     });
     renderPassDescriptor.timestampWrites = {
       querySet,
@@ -284,7 +274,17 @@ const init: SampleInit = async ({ canvas, pageState, gui }) => {
       passEncoder.end();
     }
 
-    if (hasTimestampQuery && resultBuffer.mapState == 'unmapped') {
+    let resolveBuffer: GPUBuffer | undefined = undefined;
+    let resultBuffer: GPUBuffer | undefined = undefined;
+    if (hasTimestampQuery) {
+      resolveBuffer = device.createBuffer({
+        size: 4 * BigInt64Array.BYTES_PER_ELEMENT,
+        usage: GPUBufferUsage.QUERY_RESOLVE | GPUBufferUsage.COPY_SRC,
+      });
+      resultBuffer = device.createBuffer({
+        size: 4 * BigInt64Array.BYTES_PER_ELEMENT,
+        usage: GPUBufferUsage.COPY_DST | GPUBufferUsage.MAP_READ,
+      });
       commandEncoder.resolveQuerySet(querySet, 0, 4, resolveBuffer, 0);
       commandEncoder.copyBufferToBuffer(
         resolveBuffer,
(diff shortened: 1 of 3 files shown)


beaufortfrancois:
@kainino0x Can you review?

greggman:
I had no issue with this design: 

https://jsgist.org/?src=8b8ad1e7e7cfd92b6275927b9d61de23

Which is effectively, you have an array of free queries and result buffers, when you start a frame you ask for a query/result buffer. If there are none free you allocate new ones. You then mapAsync.then() to get the result. After reading the result you add the query and result buffer back to the free list. You end up at most with 2-4 of these after a few frames and then no more allocations

You might get the result later 1-3 frames later but you get no jank

beaufortfrancois:
Thanks for sharing @greggman! Is the pattern you're recommending is the one we should promote or is it one among others? I'm genuinely asking as it's not clear to me ;)

beaufortfrancois:
I've removed `resultBuffer.mapState == 'unmapped'` check in latest patch by making sure buffers are created each frame, which work great locally.

I humbly think having a list of free buffers is a bit too much for a sample.

greggman:
>  Is the pattern you're recommending is the one we should promote or is it one among others? I'm genuinely asking as it's not clear to me ;)

I don't know either, haha. I'd like to know too.

I think having a list of buffers is trivial code. 

```
freeBuffers = [];

buffer = freeBuffers.pop() || device.createBuffer(...);

...

buffer.mapAsync().then(_ => { 
  ...
  freeBuffers.push(buffer);
});
```

src/sample/computeBoids/main.ts
```diff
@@ -231,8 +273,40 @@ const init: SampleInit = async ({ canvas, pageState, gui }) => {
...
+    }
+
     device.queue.submit([commandEncoder.finish()]);
 
+    if (hasTimestampQuery) {
+      resultBuffer.mapAsync(GPUMapMode.READ).then(() => {
+        const times = new BigInt64Array(resultBuffer.getMappedRange());
+        const renderPassDuration = Number(times[1] - times[0]);
+        const computePassDuration = Number(times[3] - times[2]);
+        perfDisplay.textContent = ` render pass duration: ${renderPassDuration}ns\ncompute pass duration: ${computePassDuration}ns\n`;
```
kainino0x, line 305: The compute pass runs before the render pass, so the order here should be flipped.

I'd also flip the indices in the query set just so they're consistent (0/1 for compute, 2/3 for render)

src/sample/computeBoids/main.ts
```diff
@@ -231,8 +273,40 @@ const init: SampleInit = async ({ canvas, pageState, gui }) => {
       passEncoder.draw(3, numParticles, 0, 0);
       passEncoder.end();
     }
+
+    let resolveBuffer: GPUBuffer | undefined = undefined;
+    let resultBuffer: GPUBuffer | undefined = undefined;
+    if (hasTimestampQuery) {
+      resolveBuffer = device.createBuffer({
```
kainino0x, line 280: There's no need to create a new resolveBuffer each frame. It never gets mapped so it's always ready to reuse in the next frame.

src/sample/computeBoids/main.ts
```diff
@@ -231,8 +273,40 @@ const init: SampleInit = async ({ canvas, pageState, gui }) => {
...
+        size: 4 * BigInt64Array.BYTES_PER_ELEMENT,
+        usage: GPUBufferUsage.QUERY_RESOLVE | GPUBufferUsage.COPY_SRC,
+      });
+      resultBuffer = device.createBuffer({
```
kainino0x, line 284: It is ideal to avoid creating a new resultBuffer every frame, and we should especially avoid it in samples, where we want to show best practices.

One way to do this would be to simply skip the copyBufferToBuffer call if resultBuffer is still mapped from a previous frame. We wouldn't get a performance result every frame, but that's OK.

That said, a ring buffer for readbacks is a very useful thing to demonstrate in a sample, and I'm not sure we have any other samples that do it, so I say let's go for it. Gregg's code is tiny and pretty simple to understand. If we do this, it would be good to show the size of the ring buffer in the little performance overlay. Hopefully it doesn't grow too large or do weird things when the tab is backgrounded or resized or anything - we should spot-check it.

src/sample/computeBoids/main.ts
beaufortfrancois: Thanks!

src/sample/computeBoids/main.ts
beaufortfrancois: Got it.

src/sample/computeBoids/main.ts
beaufortfrancois: I've implemented ring buffer for readbacks as suggested.

beaufortfrancois:
I've addressed @kainino0x and @greggman feedback in latest patch.

kainino0x (approved):
LGTM but had a few nits which I resolved in the commit I'll push now:
- Number is hard to read because it flickers too fast, and it's too big (I changed it to µs)
- Some comments would help explain to readers

kainino0x:
image

beaufortfrancois:
Thank you very much @kainino0x! It looks great

PR #325, opened 2023-11-14
Make cameras sample mobile friendly

beaufortfrancois:
This PR makes [cameras sample](https://webgpu.github.io/webgpu-samples/samples/cameras) mobile friendly by using `pointer` events instead of `mouse` desktop events when appropriate. It also makes sure those events apply only to the canvas as it currently applies to the entire window, which is confusing even on desktop.
Lastly, it fixes a typo.

FYI @ben-clayton


ben-clayton (approved):
This was on my TODO list. Thanks!

PR #327, opened 2023-11-17
Fix confusing code in deferredRendering

tyfkda:
I modified the Deferred Rendering implementation due to some confusing code:

* The initialization of `viewMatrix` incorrectly uses `mat4.inverse` with `mat4.lookAt`.
* The `getCameraViewProjMatrix` function has a local variable of the same name, so it's not used.
* `rotation` matrix is applied twice within the `getCameraViewProjMatrix` function.
* I've changed `viewProjMatrix` to not reserve space in advance for the result matrix, similar to other matrices like `rotation` that don't do this.

### DIFF
--- a/src/sample/deferredRendering/main.ts
+++ b/src/sample/deferredRendering/main.ts
@@ -487,10 +487,6 @@ const init: SampleInit = async ({ canvas, pageState, gui }) => {
     2000.0
   );
 
-  const viewMatrix = mat4.inverse(mat4.lookAt(eyePosition, origin, upVector));
-
-  const viewProjMatrix = mat4.multiply(projectionMatrix, viewMatrix);
-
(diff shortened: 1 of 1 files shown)


shrekshao (approved):
LGTM. Thanks for the change. I have no idea why the original code looks like that. Probably leftover when migrating from `gl-matrix` to `wgpu-matrix`

One comment: Remove `eyePosition` at line 479 as indicated by GithubActions
(I cannot add a comment there as github says "Line must be part of the diff")

tyfkda:
@shrekshao Thank you for your review.
I've updated (amend) my commit
and it seems passing `lint`, `build` and `export` on my local environment.

tyfkda:
@shrekshao, @austinEng,
Would you please proceed merge process? Thanks.

PR #329, opened 2023-11-21
Improve timestamp stability on Compute Boids

toji:
Improves the stability of the timestamps on the compute boids sample by rejecting any negative durations. Negative durations are expected occasionally as [described by the spec](https://gpuweb.github.io/gpuweb/#timestamp):

> Timestamp values are implementation defined and may not increase monotonically. The physical device may reset the timestamp counter occasionally, which can result in unexpected values such as negative deltas between timestamps that logically should be monotonically increasing. These instances should be rare and can safely be ignored. Applications should not be written in such a way that unexpected timestamps cause an application failure.

By discarding any negative durations and only considering positive ones the timings reported by the page become much more stable, and no longer have random drops to single-digit values as witnessed previously on some devices.

CC: @beaufortfrancois

### DIFF
--- a/src/sample/computeBoids/main.ts
+++ b/src/sample/computeBoids/main.ts
@@ -261,6 +261,7 @@ const init: SampleInit = async ({ canvas, pageState, gui }) => {
   let t = 0;
   let computePassDurationSum = 0;
   let renderPassDurationSum = 0;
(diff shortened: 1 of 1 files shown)


kainino0x (approved):
LGTM!

PR #330, opened 2023-11-22
Update perf display style

beaufortfrancois:
This PR moves the perf display to the bottom left (instead of top left) so that controls can be accessed on mobile devices. It also updates style to make it look nicer (to my eyes at least).

**old vs. new**


beaufortfrancois:
@kainino0x Please review.

cc @toji

greggman (approved):

toji (approved):
LGTM, thanks! I was going to try and tweak this today but I'm glad you got to it first because my version would have been uglier.

PR #331, opened 2023-11-22
Make GUI appear on top of other elements

greggman:
I was surprised to learn that this didn't default
to z-Index > 0. I think it works in most apps because it's expected not to call the GUI constructor until after your other elements have been created? (I didn't look too deeply)

It's possible you might want to move elements, see https://github.com/webgpu/webgpu-samples/pull/330

But in any case, it seemed like a good default.

Before:


toji (approved):
LGTM, thank you!

PR #336, opened 2023-12-07
Animometer Replace Magic Number with device.limits.minUniformBufferOffsetAllignment

cmhhelgeson:
Replaced use of 256 to calculate alignedUniformBytes with a constant holding the value of device.limits.minUniformBufferOffsetAlignment. Ideally this serves a few goals.

1. Saving space in the hypothetical scenario where minUniformBufferOffsetAlignment becomes smaller.
2. Adding a critical piece of context to the calculation of alignedUniformBytes that might not be immediately clear to a user that hasn't worked with buffer offsets in WebGPU before (i.e the size of the offsett is based around the device's limits, and is made to align within those limits)

I'm not an expert at using buffer offsets in WebGPU, so please inform me if my understanding of the use of this value is incorrect.


greggman:
`minUniformBufferOffsetAllignment` does not change on it's own. It only changes if you request it to be changed. Otherwise, it's effectively a constant and no no need to look it up

cmhhelgeson:
Understood. I think for learning purposes it might be helpful to specify it explicitly, but I defer to you.

PR #340, opened 2023-12-14
Simplified the compute boids example

martinlaxenaire:
This is a proposal to simplify the compute boids example by removing the buffers/bind groups ping pong.
Instead we just create one particle buffer and one bind group. Since the particle storage has a `read_write` access, there's no need to ping pong between two buffers.

I don't know why it was done like that in the first place but I suppose the storages access/capabilities might have changed since then?
Anyway I think that way it would be easier to understand and reproduce in another context.

Thanks for all the hard work by the way!

### DIFF
--- a/src/sample/computeBoids/main.ts
+++ b/src/sample/computeBoids/main.ts
@@ -215,51 +215,80 @@ const init: SampleInit = async ({ canvas, pageState, gui }) => {
     initialParticleData[4 * i + 3] = 2 * (Math.random() - 0.5) * 0.1;
   }
 
-  const particleBuffers: GPUBuffer[] = new Array(2);
-  const particleBindGroups: GPUBindGroup[] = new Array(2);
-  for (let i = 0; i < 2; ++i) {
-    particleBuffers[i] = device.createBuffer({
-      size: initialParticleData.byteLength,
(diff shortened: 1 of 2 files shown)


ben-clayton:
Because you're now reading and writing to the same storage buffer locations across invocations without barriers, you've introduced data races, and the reads can now produce undefined values.
This means the reads could give you the value before the write, the value after the write, or just complete garbage.

If the invocations of each workgroup only performed reads / writes to memory locations exclusively used by that workgroup, then you could fix this data race with a [storage barrier](https://www.w3.org/TR/WGSL/#storageBarrier-builtin). However, this sample dispatches [multiple workgroups](https://github.com/webgpu/webgpu-samples/blob/6397f232571e626ef312b2b39ba01e28ff1e6f5b/src/sample/computeBoids/main.ts#L281), and each workgroup reads from the memory written to by other workgroups, so you can't use a barrier here to fix the race.

For more information, [see this thread](https://stackoverflow.com/questions/72035548/what-does-storagebarrier-in-webgpu-actually-do)

martinlaxenaire:
Oh, damn, I understand now, thanks for the explanation.

Closing the PR.

PR #343, opened 2024-01-02
Scrollbar for Samples with Numerous Sources

cmhhelgeson:
Added an overflow scrollbar for samples with numerous code sources, such as the deferred rendering, Cornell box, or bitonicSort samples. 
https://github.com/webgpu/webgpu-samples/assets/62450112/0674c338-7929-4cc1-92f4-9524868fed2f
Draft is currently in PR because I would like to apply analogous Firefox styling for viewing in Nightly Firefox builds.


cmhhelgeson:
Actually, applying scrollbar-width: thin seems to be close enough for now.

austinEng (approved):

kainino0x:
Maybe this makes it a bit hard to see that there are more tabs to the right? Right now we flow the tabs onto a new line so they're easy to see.

cmhhelgeson:
> Maybe this makes it a bit hard to see that there are more tabs to the right? Right now we flow the tabs onto a new line so they're easy to see.

Possibly, but I'd contend that that's only likely if the end of one code "tab" aligns perfectly with the edge of the screen. I think the presence of the scrollbar and a code tab's title going off-screen is enough to clue in the user. Of course, this is a non-essential, subjective change, so it's ultimately just a matter of preference between the stacked and scrolling versions.

kainino0x:
Thinking about this further, if someone doesn't have a touchpad and doesn't know shift-scroll then horizontal scrolling is a bit painful. I would prefer a larger scrollbar (it would be more obvious, and easier to click), though I suppose I'm still not a fan of horizontal scrolling in general.

With a larger scrollbar though, I'd be happy enough either way. Thanks for putting up with my opinions :)

PR #344, opened 2024-01-04
CSS: Animated Resizing of .panelContents when Window Size is less than 768 pixels.

cmhhelgeson:
https://github.com/webgpu/webgpu-samples/assets/62450112/fe4619d4-deec-481b-8475-136e2a32c56e

In order for this animation to be performant, I replaced the React state update with direct updates to the DOM based on window resizing and clicks, which prevents the state-triggered re-render that would stall the animation/transition. Since I'm unsure whether updating the DOM in this manner would be considered a "brittle" or "unidiomatic" practice, and since it modifies a longstanding component of this repository's code, I've tried to do my best to make the change as simple and unobtrusive as possible, as well as adding comments to the new window resize listener and stylePanel functions.


austinEng:
indeed it is risky to muck directly with the DOM because it could interfere with React. Would a pure CSS solution be possible?

haven't tested it, but perhaps something like:

```css
.panel .panelContents {
   display: block;
   transition: max-height 0.25s ease-in;
 }
 
.panel[data-expanded=false] .panelContents {
    max-height: 0vh;
}

.panel[data-expanded=true] .panelContents {
  max-height: 100vh;
}
```

cmhhelgeson:
> indeed it is risky to muck directly with the DOM because it could interfere with React. Would a pure CSS solution be possible?
> 
> haven't tested it, but perhaps something like:
> 
> ```css
> .panel .panelContents {
>    display: block;
>    transition: max-height 0.25s ease-in;
>  }
>  
> .panel[data-expanded=false] .panelContents {
>     max-height: 0vh;
> }
> 
> .panel[data-expanded=true] .panelContents {
>   max-height: 100vh;
> }
> ```

This was similar to a solution I tried earlier, using the current data-expanded state object to trigger the transition. However, the issue with that approach was that the state change seemed to trigger a re-render that made the animation imperceptible. I do agree with you that most of this can probably be changed to css, perhaps with the vanilla Javascript only containing a call to setAttribute for data-expanded rather than relying on React state. I'll work on that.

cmhhelgeson:
Should be good now, I cut down the Javascript to just setting the attribute on the panel ref.

cmhhelgeson:
Just to reiterate, setAttribute is called rather than tying data-expanded to a React state variable because setting a state variable causes the entire component to re-render, including the canvas that's displaying the WebGPU code. You can actually see this problem on the current version of the website, where clicking the expand button on the cornell sample will stall the webpage for several seconds.

austinEng:
Sorry for taking a while to look at this again.

I don't feel comfortable mixing the DOM attribute manipulations while also using React here. Not an expert on this at all, but I think React provides facilities to say that things should not rerender. I believe [React.memo](https://react.dev/reference/react/memo) helps with this.

cmhhelgeson:
> Sorry for taking a while to look at this again.
> 
> I don't feel comfortable mixing the DOM attribute manipulations while also using React here. Not an expert on this at all, but I think React provides facilities to say that things should not rerender. I believe [React.memo](https://react.dev/reference/react/memo) helps with this.

Not sure if this is good practice, but I figured memoizing the component was the easiest way to fix the issue.

austinEng (approved):
Nice! I think it's great. I verified with the React devtools profiler that we're not seeing any rerenders of that Component now

cmhhelgeson:
> Nice! I think it's great. I verified with the React devtools profiler that we're not seeing any rerenders of that Component now

Glad to hear, think I just overcomplicated it earlier😅, but I think it's a good visual/usability fix.

PR #347, opened 2024-01-19
Add Timestamp Queries to Bitonic Sort Example

cmhhelgeson:
Utilized example code found at https://developer.chrome.com/blog/new-in-webgpu-121 to add timestamp queries for the bitonic sort on compatible browsers. 

At large enough element sizes, these queries can be used to compare the performance between sorts of various sizes, or compare performance when the maximum invocations per dispatch are arbitrarily limited.

Minor changes:
'Complete Sort' has been renamed to 'Auto Sort' to better reflect controller functionality.


src/sample/bitonicSort/main.ts
```diff
@@ -65,9 +67,35 @@ const getNumSteps = (numElements: number) => {
...
+    context,
+    canvas,
+    timestampQueryAvailable,
+  }) => {
     const maxInvocationsX = device.limits.maxComputeWorkgroupSizeX;
 
+    let querySet: GPUQuerySet;
+    let timestampQueryResolveBuffer: GPUBuffer;
+    let timestampQueryResultBuffer: GPUBuffer;
+    console.log(timestampQueryAvailable);
```
austinEng, line 84:
```suggestion
```
prefer not to log things automatically

src/sample/bitonicSort/main.ts
```diff
@@ -145,17 +173,19 @@ SampleInitFactoryWebGPU(
...
-      'Sort Speed': 50,
-
+      // The speed at which each step of the bitonic sort will be executed after 'Auto Sort' has been called.
+      'Auto Sort Speed': 50,
       // MISCELLANEOUS SETTINGS
       'Display Mode': 'Elements',
       // An atomic value representing the total number of swap operations executed over the course of the bitonic sort.
       'Total Swaps': 0,
+      // Time taken to sort in seconds (not nanoseconds!). If timestep query feature is not available then this will remain 0.
+      'Sort Time': '0s',
```
austinEng, line 187: why not display it in.. milliseconds? right now the seconds are hard to read. For me it sometimes says things like "0.000065536s"

austinEng (approved):
observation: it's interesting that on my machine, the first sort step takes 0.06ms and the last one takes 2ms

cmhhelgeson:
> observation: it's interesting that on my machine, the first sort step takes 0.06ms and the last one takes 2ms

Maybe it's not clear from my comments yet, but the displayed time is accumulating the times of each step rather than displaying the time a single step takes to execute. Given that the last step is almost certainly swapping over global indices rather than workgroup local indices at greater than 512 elements, it stands to reason that the last step would take longer than the first step, but there wouldn't be such a great difference. Perhaps there should be a display controller that displays the time a single step takes to execute (i.e if you're using Execute Sort Step rather than Auto Sort), a display controller that accumulates the times for each step within a sort, and maybe a final display controller that averages out the time a sort takes under a certain sort config combination (Elements + Workgroup Size Limit)

cmhhelgeson:
I think I'll fix/add that (maybe move this back to draft or ask for re-review later).

cmhhelgeson:
> I think I'll fix/add that (maybe move this back to draft or ask for re-review later).

This should be finished now bar any unexpected bugs.

austinEng (approved):

PR #349, opened 2024-01-24
Skinned Mesh Example

cmhhelgeson:
A port of the WebGLFundamentals skinned mesh tutorial to webgpu-samples.

This is what the skinned grid currently looks like.

Skinned Grid

And this is what the whale currently looks like:
Whale

I'm going to try to cut out a lot of the GLTF class cruft to get to the minimal viable version of this without having to deal with all the GLTF abstraction nonsense, but any additional advice would be appreciated.


austinEng:
> I'm going to try to cut out a lot of the GLTF class cruft to get to the minimal viable version of this without having to deal with all the GLTF abstraction nonsense, but any additional advice would be appreciated.

@toji @greggman might have suggestions about this

cmhhelgeson:
> > I'm going to try to cut out a lot of the GLTF class cruft to get to the minimal viable version of this without having to deal with all the GLTF abstraction nonsense, but any additional advice would be appreciated.
> 
> @toji @greggman might have suggestions about this

Just wanted to heads-up that I probably won't be able to get to this today. But I'm going to try working with the simpleSkin GLTF example first to get that working before moving to the whale.

toji:
Having implemented glTF skinning in WebGPU a couple of times now (in https://github.com/toji/burrow and https://github.com/toji/spookyball) and not being particularly satisfied with either of them, I'm not sure I have a lot of great advice because it's just kind of complicated no matter how you deal with it.

I guess one thing I would highly recommend is to not try to make a general purpose glTF loader, as those can get complicated fast, and dealing with all the edge cases is painful. I'd pick a mesh or two you want to render, push them through something like [gltf-transform](https://gltf-transform.dev/) to make sure they're nicely normalized, and just focus on what you need to load that. Things like assuming that you'll always get a specific set of attributes in a specific format can significantly simplify the loading. Plaster the sample and the code with comments plainly stating that you don't intend it to be a robust glTF loader, so that people don't complain when it doesn't work on their mesh. (Or, more accurately: _when_ they complain you can point them at the comments and recommend a more robust library.)

Also, these days compute skinning is all the rage, and WebGPU is perfectly capable of doing it (See: https://github.com/toji/burrow/blob/main/src/renderer/render-utils/compute-skinning.ts) but it introduces a lot of additional compute-specific complications (see: https://toji.dev/webgpu-best-practices/compute-vertex-data) and so I wouldn't recommend trying to demonstrate it here. At least not as a first step.

cmhhelgeson:
> Having implemented glTF skinning in WebGPU a couple of times now (in https://github.com/toji/burrow and https://github.com/toji/spookyball) and not being particularly satisfied with either of them, I'm not sure I have a lot of great advice because it's just kind of complicated no matter how you deal with it.
> 
> I guess one thing I would highly recommend is to not try to make a general purpose glTF loader, as those can get complicated fast, and dealing with all the edge cases is painful. I'd pick a mesh or two you want to render, push them through something like [gltf-transform](https://gltf-transform.dev/) to make sure they're nicely normalized, and just focus on what you need to load that. Things like assuming that you'll always get a specific set of attributes in a specific format can significantly simplify the loading. Plaster the sample and the code with comments plainly stating that you don't intend it to be a robust glTF loader, so that people don't complain when it doesn't work on their mesh. (Or, more accurately: _when_ they complain you can point them at the comments and recommend a more robust library.)
> 
> Also, these days compute skinning is all the rage, and WebGPU is perfectly capable of doing it (See: https://github.com/toji/burrow/blob/main/src/renderer/render-utils/compute-skinning.ts) but it introduces a lot of additional compute-specific complications (see: https://toji.dev/webgpu-best-practices/compute-vertex-data) and so I wouldn't recommend trying to demonstrate it here. At least not as a first step.

Definitely not intending to make a general purpose loader, but I didn't know about the compute skinning, so maybe I'll look into that later. 

I am finding a lot of pretty obvious bugs in the code, so maybe I'm closer than I think.

cmhhelgeson:
@toji Are the alignment issues with vec3<f32>s described in this article: https://toji.dev/webgpu-best-practices/compute-vertex-data no longer a concern? I tried writing a simple misaligned passThrough shader in a separate branch and it seemed to work fine?

```
struct VertexUniforms {
  count: u32,
}
@group(0) @binding(0) var<uniform> vertex_uniforms: VertexUniforms;
@group(0) @binding(1) var<storage, read_write> positions: array<vec3<f32>>;

@compute @workgroup_size(64)
fn passThrough(@builtin(global_invocation_id) global_id : vec3<u32>) {
  let index = global_id.x;
  // Return early if the vertex index is higher than half the number of vertices
  if (index >= vertex_uniforms.count) { 
    return; 
  }
  let current_position = positions[index];
  positions[index] = current_position * 1.00001;
}
```

cmhhelgeson:
Probably too wobbly for general use, but I believe, barring any further suggestions, that this is now working as it is intended to.

https://github.com/webgpu/webgpu-samples/assets/62450112/b533a676-8249-4c1b-9448-9b1afb5493df

src/sample/skinnedMesh/gltf.wgsl
```diff
@@ -0,0 +1,74 @@
...
+@group(3) @binding(0) var<uniform> joint_matrices: array<mat4x4<f32>, 6>;
+
+@vertex
+fn vertexMain(input: VertexInput) -> VertexOutput {
+  var output: VertexOutput;
+  let skin_matrix = 
+    joint_matrices[input.joints[0]] * input.weights[0] +
+    joint_matrices[input.joints[1]] * input.weights[1] +
+    joint_matrices[input.joints[2]] * input.weights[2] +
+    joint_matrices[input.joints[3]] * input.weights[3];
```
toji, line 46: You're not taking the inverse bind matrices into account here, which probably accounts for some of the strange "droopy" look of the skinning you were talking about. See https://github.com/toji/spookyball/blob/main/js/engine/webgpu/wgsl/common.js#L49 for an example of some of my shader code that incorporates the IBM.

cmhhelgeson: Aren't they taken into account when whaleScene.skins[0].update is called each frame, and the inverseBindMatrix times the worldMatrix is passed into the inverseBindMatricesUniformBuffer (which corresponds to the joint_matrices var<uniform> in the shader)?

cmhhelgeson: I.E that particular step of the calculation is happening on the CPU rather than the GPU on lines 716-739 of glbUtils.ts

src/sample/skinnedMesh/main.ts
```diff
@@ -0,0 +1,527 @@
...
+    for (let i = 0; i < skin.joints.length; i++) {
+      const joint = skin.joints[i];
+      if (!origMatrices.has(joint)) {
+        origMatrices.set(joint, whaleScene.nodes[joint].source.getMatrix())
+      }
+      const origMatrix = origMatrices.get(joint);
+      const m = mat4.rotateX(origMatrix, angle);
+      whaleScene.nodes[joint].source.position = mat4.getTranslation(m);
+      // Something wrong with scaling
+      //whaleScene.nodes[joint].source.position = mat4.getScaling(m);
```
toji, line 376: I would encourage you to figure out why this wasn't working rather than just commenting it out. People *will* be referring to the code as a first step for implementing their own skinning systems, so it's worth making it as solid as we reasonably can.

src/sample/skinnedMesh/main.ts
```diff
@@ -0,0 +1,527 @@
...
+    const modelMatrix = getModelMatrix();
+
+    // Calculate bone transformation
+    const t = Date.now() / 20000 * settings.speed;
+    const angle = Math.sin(t) * settings.angle;
+    // Compute Transforms when angle is applied
+    computeBoneMatrices(gridBoneCollection.transforms, angle);
+    gridBoneCollection.transforms.forEach((boneTransform, idx) => {
+      // Apply inverseBindPose to normal transform to get transform passed to our uniforms
+      mat4.multiply(boneTransform, gridBoneCollection.bindPosesInv[idx], gridBoneCollection.uniforms[idx])
```
toji, line 399: You ARE applying the inverse bind pose here, but that seems like it's only for the grid mesh? I'd recommend combining the joint transforms and the inverse bind transforms in the shader for both meshes. It's more GPU work and slightly more GPU memory, yes, but it also means less per-frame JavaScript work which is going to be favorable for most WebGPU pages.

src/sample/skinnedMesh/glbUtils.ts
```diff
@@ -0,0 +1,946 @@
...
+        },
+      ],
+    });
+  }
+
+  update(device: GPUDevice, node: GLTFNode, nodes: GLTFNode[]) {
+    const globalWorldInverse = mat4.inverse(node.worldMatrix);
+    for (let j = 0; j < this.joints.length; j++) {
+      const joint = this.joints[j];
+      const dstMatrix: Mat4 = this.jointMatrices[j];
```
toji, line 718: Is this a copy, or a view into the original jointMatrices storage? If it's a view, then doesn't it get overwritten every frame by the `mat4.multiply(dstMatrix, ....);` below?

src/sample/skinnedMesh/gltf.wgsl
toji: I didn't see that, sorry. Though I'm still a little concerned about how that's working at the moment? I left a separate comment.

src/sample/skinnedMesh/main.ts
cmhhelgeson: Can't believe I didn't see this, but as you can see, the scaling was applied to position...

src/sample/skinnedMesh/glbUtils.ts
cmhhelgeson: Turns out, jointMatrices wasn't really needed, as dstMatrix just acts as a storage array for what will be written to the uniformBuffer, and doesn't need to have the previous frame's value accessible for any reason.

src/sample/skinnedMesh/main.ts
toji: Oh! I didn't notice that either! Easy mistake to make, and I'm glad you found it!

toji:
I loaded this up locally and, welp...

https://github.com/webgpu/webgpu-samples/assets/805273/59a788a4-47f0-43a3-9591-d7fc416c2456

As much as I honestly adore that little dance, I don't think it's quite what's intended here. 🐋🕺

I'll look through the code and see if I can figure anything else out. Thanks for your ongoing work on this!

cmhhelgeson:
> I loaded this up locally and, welp...
> 
>  Skinned.Mesh.-.WebGPU.Samples.-.Trim.mp4 
> As much as I honestly adore that little dance, I don't think it's quite what's intended here. 🐋🕺
> 
> I'll look through the code and see if I can figure anything else out. Thanks for your ongoing work on this!

Oh, I just changed the guy to rotate on the z axis instead of the x-axis because I thought it looked cuter.

src/sample/skinnedMesh/gltf.wgsl
```diff
@@ -0,0 +1,98 @@
...
+// Ways this shader can be made more generalizable include:
+//    a. Making our shader a constructable string returned from a function that takes the skin's current number of joints as an argument.
+//       For example, converting array<mat4x4f, 6> to return `array<mat4x4f, ${numJoints}`
+//    b. Converting our uniform matrices buffers into storage buffers of variable size
+//       joint_matrices: var<uniform> joint_matrices -> var<storage, read> joint_matrices
+//    c. Reading our matrix data in as a texture
+// However, for this limited, single object example, the current approach works fine, and these considerations are only
+// necessary in the context of a more robust gltf parser/shader package.
+@group(3) @binding(0) var<uniform> joint_matrices: array<mat4x4<f32>, 6>;
+@group(3) @binding(1) var<uniform> inverse_bind_matrices: array<mat4x4<f32>, 6>;
```
toji, line 48: I just noticed while looking through the code that you're doing a fixed joint array size here. I really appreciate the comment explaining the reasoning behind this, but I would advocate for making the change described in option "b" here: Change the joint arrays to storage buffers (which can subsequently have a runtime size array.)

While I recognize that the existing code probably lines up better with the equivalent WebGLFundamentals shader, we should generally aim to demonstrate good use of the WebGPU API and I think the most general case for developers would be to use storage here.

Fortunately it's a very minor change, as you should only have to change the signature here and the bind group layout. Everything else should be able to stay exactly as-is!

cmhhelgeson: I changed this as requested, but unfortunately, I still believe there is an issue with the skinning. For some reason the middle of the body does not appear to animate when animSkin is called, although the front and end of the body do (note how the final skinning example in webgl2fundamentals has the whale arch its back, whereas in the current code, the whale just droops its head down).

cmhhelgeson: Uploading Skinned Mesh - WebGPU Samples - Google Chrome 2024-01-30 15-50-45.mp4…

cmhhelgeson: It appears that each joint _is applying_ an animation (i.e when you force the code to only animate one joint rather than all of them, there is palpable movement in a differing area).

cmhhelgeson: Looking at the original source code, it seems as if the model used in that example might actually have a different joint structure, as its joints are listed as 6, 7, 8, 9, 10, 11 rather than our joints which reference nodes 5, 2, 1, 0, 3, and 4. Perhaps the mesh has changed, or Blender's builtin GLTF exporter works differently than the exporter used in the tutorial.

toji: I wouldn't worry about the differences in joint numbers. That could easily be differences in export methods like you mentioned.

To verify you can always load your version of the model into a known-good viewer like https://github.khronos.org/glTF-Sample-Viewer-Release/ to make sure that the animation looks as expected.

toji: Actually, a follow up to that: there IS at least one place where your code hard-codes a specific node number (6, If I recall). If that was based on the nodes from the WebGLFundamentals version of the model then you might need to adjust it.

cmhhelgeson: Hard-coding node 6 is based on the raw json output, (node 6 has mesh 0 and skin 0).

I'm not sure if I can use the gltf viewer to bug test the animation, as the animation is not driven by the loaded gltf animations, but by code within main.ts that applies new position, rotation, and scale values to the BaseTransform of the given node.

cmhhelgeson:
At the moment, I think the issues have more to do with the animation rather than the skinning of the mesh itself, as each node seems to properly align with the proper joint and such. Accordingly, I've changed the animation a little bit to be more natural, as I think that might have been the real issue (ie when everything joint was rotating the same amount at the same time, the movement of the mesh did not seem "natural").

cmhhelgeson:
Hello, just wanted to check and see if this needs more work.

toji:
I apologize for disappearing for a bit there. I just got back from a family vacation, so I'm catching up on various things.

>I'm not sure if I can use the gltf viewer to bug test the animation, as the animation is not driven by the loaded gltf animations, but by code within main.ts that applies new position, rotation, and scale values to the BaseTransform of the given node.

I didn't realize initially that you were manually moving the joints in code and not parsing animations from the file, and I apologize for that because it likely made for some confusing review comments. While I'm a little torn on this I think ultimately avoiding the complexity of parsing the glTF animations is the right call for an example that wants to keep the focus on the technique of skinning.

With that in mind, I think it's likely fine, but give me a little bit longer to take one more pass through the code. I'll try to have that to you by end of day today or early tomorrow.

public/assets/gltf/RiggedSimple.glb
toji, line 1: This file isn't used in the example, so it should be removed.

public/assets/gltf/simpleSkin.glb
toji, line 1: Same here, this file doesn't appear to be used.

src/sample/skinnedMesh/grid.wgsl
toji, line 1: I don't think this needs to be solved before this lands, but at this point this file is so similar to `gltf.wgsl` that it would be really nice if we could combine the shaders somehow. A future update that allows them to be combined somehow would be nice (either by having some sort of preprocessor pass to include things like the normal math only when needed or by simply providing the grid with dummy normals.)

src/sample/skinnedMesh/main.ts
toji, line 1: I would prefer to see this file have more comments. Since these examples are meant to be educational, anything that would help someone understand what is happening and why is beneficial.

In terms of audience, I think it's reasonable to assume that the reader has some knowledge of skinning as a concept (because there's lots of resources available that explain it in depth) and are most likely looking to this example to figure out how to execute that concept in WebGPU specifically.

src/sample/skinnedMesh/main.ts
```diff
@@ -0,0 +1,570 @@
...
+    requestAnimationFrame(frame);
+  }
+  requestAnimationFrame(frame);
+};
+
+const skinnedMesh: () => JSX.Element = () =>
+  makeSample({
+    name: 'Skinned Mesh',
+    description:
+      'A demonstration of basic gltf loading and mesh skinning, ported from https://webgl2fundamentals.org/webgl/lessons/webgl-skinning.html. Mesh data, per vertex attributes, and skin inverseBindMatrices are taken from the json parsed from the binary output of the .glb file, with animated joint matrices updated and passed to shaders per frame via uniform buffers.',
```
toji, line 529: To avoid having readers make the same mistake I did, maybe update this to say something like:

```suggestion
      'A demonstration of basic gltf loading and mesh skinning, ported from https://webgl2fundamentals.org/webgl/lessons/webgl-skinning.html. Mesh data, per vertex attributes, and skin inverseBindMatrices are taken from the json parsed from the binary output of the .glb file. Animations are generated programmatically with animated joint matrices updated and passed to shaders per frame via uniform buffers.',
```

toji (commented):
Overall I think this is in a pretty good place! Aside from a couple of extremely minor nits the biggest thing I'd like to see is some additional comments to help explain the code to developers, then I think we can land this!

public/assets/gltf/RiggedSimple.glb
cmhhelgeson: Deleted

src/sample/skinnedMesh/main.ts
cmhhelgeson: I added some more comments, though I'm not sure if there should be more.

src/sample/skinnedMesh/main.ts
cmhhelgeson: Changed

toji (approved):
Thank you for the updates and for your persistence in getting this landed. Looks good to me!

ben-clayton:
I don't believe the PR author has permissions to land, so merging on their behalf.

PR #351, opened 2024-01-26
Improper vs Proper Vertex Allignment when Writing to Vertex Buffer From Compute Shader

cmhhelgeson:
Not sure if this would be too niche or low density of an example, but inspired by Toji's article (https://toji.dev/webgpu-best-practices/compute-vertex-data) on handling vertex data in compute shaders, I've created this simple sample showing how shaders that seemingly perform the same result can fail to execute depending on how vertex data is interpreted by the compute shader.

The example is mostly done, with tweaks only really needed for additional comments, explanations, or questions about the validity of the example as a test case when compared to other samples which much more comprehensively exhibit WebGPU's capabilities rather than addressing a small consideration one needs to take when handling vertex data in compute shaders.


austinEng:
Thanks for writing this!

I do think it's a little too niche to be a "sample". I think it would be great to have something explaining and/or showing this alignment issue, perhaps on https://webgpufundamentals.org/ if there isn't something already. WebGPU fundamentals is also in this Github org at https://github.com/webgpu/webgpufundamentals. @greggman maintains it - what do you think?

cmhhelgeson:
> Thanks for writing this!
> 
> I do think it's a little too niche to be a "sample". I think it would be great to have something explaining and/or showing this alignment issue, perhaps on https://webgpufundamentals.org/ if there isn't something already. WebGPU fundamentals is also in this Github org at https://github.com/webgpu/webgpufundamentals. @greggman maintains it - what do you think?

I'd also agree it's too niche for it's own example, even if it would be more germane for a website like https://webglsamples.org/. It could possibly be folded into one of the existing samples. For instance, an option could be added to Rotating Cube to modify the vertex positions in either the vertex or the compute shader, demonstrating how vec3fs need to be handled differently depending on the shader context. However, that might obfuscate or distract from the original intention of that shader, or any other shader such an option might be applied to.

austinEng:
> As for the webgpufundamentals suggestion, a section could probably be added to https://webgpufundamentals.org/webgpu/lessons/webgpu-vertex-buffers.html, potentially containing a redirect to https://toji.dev/webgpu-best-practices/compute-vertex-data.

Sounds great to me. I don't think the current format of webgpu-samples is suitable right now to explain this alignment detail

PR #354, opened 2024-02-21
Moved duplicated utils to separate file.

cmhhelgeson:
Moved createBindGroupCluster and other util functions to common utils file to prevent code duplication. Re-documented code to make the functions more understandable/extensible to other use cases.


sample/sampleUtils.ts
```diff
@@ -0,0 +1,307 @@
...
+
+  return {
+    bindGroups,
+    bindGroupLayout,
+  };
+};
+
+export type SampleInitParams = {
+  canvas: HTMLCanvasElement;
+  pageState: { active: boolean };
```
kainino0x, line 121: all mentions of pageState are obsolete

sample/sampleUtils.ts
```diff
@@ -0,0 +1,307 @@
...
+ *   ],
+ *   resourceLayouts: [
+ *     [uniformBuffer, sampledTextureView],
+ *     [uniformBuffer2, sampledTextureView2]
+ *   ]
+ * });
+ */
+export const createBindGroupCluster = (
+  args: BindGroupClusterFunctionArgs
+): BindGroupCluster => {
```
kainino0x, line 80: This is a handy helper, but I'm a bit hesitant to hide away details of WebGPU API usage behind helpers because it adds some load to reading the samples - even if you know WebGPU, and you want to see how this technique works, you end up having to learn what the utility functions do (and this utility function is not very easy to read). Generally the style practices I'd apply to samples are a little different than normal: more focus on teaching, less on conciseness and refactorability.

I couldn't find any instances of this helper even being used with more than one bind group per cluster, though it seems on the latest revision some of the samples are broken / have broken imports\*, so I might be wrong about that. But at least in that case the amount of code needed to use the API directly is barely more than using the helper and it diminishes the value of the automatic labeling and numbering.

\* the build really shouldn't be passing, I'm getting lots of errors when I run `npm run build` on this PR, but it exits with return code 0. cc @greggman

sample/sampleUtils.ts
```diff
@@ -0,0 +1,307 @@
...
+      context,
+      presentationFormat,
+      stats,
+      timestampQueryAvailable,
+    });
+  };
+  return init;
+};
+
+// A class that provides a set of utilities for creating a basic fullscreen shader that renders
```
kainino0x, line 188: doc comments for all these comments too

sample/sampleUtils.ts
```diff
@@ -0,0 +1,307 @@
...
+// a programatically generated 2D image to the screen
+export abstract class BaseFullscreenShaderClass {
+  // The intended implementation of switchBindGroup is to facilitate switching between bind groups that conform
+  // to the same bind group layout used within the fullscreen shader
+  abstract switchBindGroup(name: string): void;
```
kainino0x, line 193: seems unused?

sample/sampleUtils.ts
```diff
@@ -0,0 +1,307 @@
...
+      stats,
+      timestampQueryAvailable,
+    });
+  };
+  return init;
+};
+
+// A class that provides a set of utilities for creating a basic fullscreen shader that renders
+// a programatically generated 2D image to the screen
+export abstract class BaseFullscreenShaderClass {
```
kainino0x, line 190: broadly speaking I think a fullscreen quad helper could make more sense for samples, depending on the implementation; in principle it's pretty straightforward to a reader what it does just from the name, and it also encapsulates more stuff.

But as written its interface seems quite complicated, especially using inheritance where I don't think it really helps much. Nothing here even uses `this` so it doesn't seem like it's using the fact that it's a class. I have left a few additional thoughts below but let me know what you think.

sample/sampleUtils.ts
```diff
@@ -0,0 +1,307 @@
...
+    bindGroups: GPUBindGroup[]
+  ) {
+    const passEncoder = commandEncoder.beginRenderPass(renderPassDescriptor);
+    passEncoder.setPipeline(pipeline);
+    for (let i = 0; i < bindGroups.length; i++) {
+      passEncoder.setBindGroup(i, bindGroups[i]);
+    }
+    passEncoder.draw(6, 1, 0, 0);
+    passEncoder.end();
+  }
```
kainino0x, line 224: IMO could be a standalone or static function or more likely just don't need a helper for this.

sample/sampleUtils.ts
```diff
@@ -0,0 +1,307 @@
...
+    keys: K
+  ) {
+    for (let i = 0; i < keys.length; i++) {
+      device.queue.writeBuffer(
+        uniformBuffer,
+        i * 4,
+        new Float32Array([instance[keys[i]]])
+      );
+    }
+  }
```
kainino0x, line 239: Doesn't seem at all specific to rendering a quad. Unclear how the `instance`/`keys` thing helps, I can't find any examples of it being used.

Also this does a lot of writeBuffer calls where just one would be much more ideal.

kainino0x (commented):
I've left a bunch of thoughts but apologies, I don't think I can spend a lot more time on reviewing this. If we're going to have shared sample utils I'd like to hold them to a high standard of general usefulness, but these utils are rather ad hoc for these particular samples. Any utils will need a design that enhances readability/teachability overall, which means it needs to overcome the (IMO) extra readability overhead that comes with trying to abstract out certain helpers. I don't think these helpers are likely to do that as they mostly hide WebGPU complexity (which is valuable to teach) behind util complexity (which is much less valuable to teach).

Issue #357, opened 2024-02-27
Organize the samples list


kainino0x:
The samples list is slowly growing and I think it needs to be divided up into a few categories (and maybe we should have a directory page that shows the descriptions of each sample, and/or topic tags on each sample in the listing). Here's a proposed categorization:

- WebGPU Demos
    - particles
    - cornell
    - computeBoids
    - gameOfLife
- WebGPU Basics
    - helloTriangle
    - helloTriangleMSAA
    - rotatingCube
    - twoCubes // consider renaming for clarity
    - texturedCube
    - instancedCube
    - fractalCube
    - samplerParameters
    - cubemap
    - normalMap
    - #127 3d textures (or could be in a larger demo under Graphics Technique Implementations)
    - #150 (same)
- Important Graphics Features
    - renderBundles
    - reversedZ
- Important Compute Features
    - #350
- Web Platform Features
    - resizeCanvas
    - #356
    - #127 canvas readback
    - videoUploading
    - videoUploadingWebCodecs
    - worker
- Compute Technique Implementations
    - bitonicSort
- Graphics Technique Implementations
    - cameras
    - #45
    - imageBlur
    - #217
    - shadowMapping
    - deferredRendering
    - A-buffer
    - skinnedMesh
- Benchmarks
    - animometer
    - Workload Simulator

kainino0x:
Another proposed version based on @cmhhelgeson feedback

- WebGPU Basics
    - helloTriangle
    - helloTriangleMSAA
    - rotatingCube
    - twoCubes // consider renaming for clarity
    - texturedCube
    - instancedCube
    - fractalCube
    - samplerParameters
    - cubemap
    - #127 3d textures (or could be in a larger demo in Techniques)
    - #150 (same)
- Important Graphics Features
    - renderBundles
    - reversedZ
- Important Compute Features
    - #350
- Web Platform Features
    - resizeCanvas
    - #356
    - #127 canvas readback
    - videoUploading
    - videoUploadingWebCodecs
    - worker
- GPGPU Techniques
    - bitonicSort
    - gameOfLife
- Graphics Techniques w/ Compute
    - imageBlur
    - A-buffer
    - particles
    - cornell
    - computeBoids
- Classic Graphics Techniques
    - cameras
    - deferredRendering
    - normalMap
    - shadowMapping
    - skinnedMesh
    - #45
    - #217
- Benchmarks
    - animometer
    - Workload Simulator

[...]

PR #358, opened 2024-02-28
Organize Sample List

cmhhelgeson:
Refer to #357 

image


cmhhelgeson:
This is ready for review, albeit I imagine there will be a lot of discussion around implementation, styling, whether categories should be collapsable, and how various samples should be collected and categorized.

src/components/SampleCategory.module.css
```diff
@@ -0,0 +1,32 @@
...
+  margin-left: 8px;
+  margin-top: 7px;
+  margin-bottom: 5px;
+}
+
+li.selected a {
+  color: #ff0000;
+}
+
+.dropdown[data-collapsed=false] {
```
kainino0x, line 26: I don't think the sections should be collapsible. The list is not going to get _that_ big and it's easier to find things if the whole list is visible (and ctrl-F-able). Examples:
https://webglsamples.org/WebGL2Samples/
https://threejs.org/examples/

src/components/SampleCategory.module.css
```diff
@@ -0,0 +1,32 @@
...
+  cursor: pointer;
+}
+
+.dropdown {
+  margin-left: 8px;
+  margin-top: 7px;
+  margin-bottom: 5px;
+}
+
+li.selected a {
```
kainino0x, line 22: Categories can't be selected, so this should be unused?

kainino0x, line 22: Categories can't be selected, so this should be unused?

src/pages/MainLayout.module.css
```diff
@@ -67,12 +73,13 @@
   .panel .panelContents {
     display: block;
     transition: max-height 0.3s ease-out;
-    overflow: hidden;
+    overflow: none;
```
kainino0x, line 76: What does this do? MDN doesn't list it as a valid value for `overflow`.

src/components/SampleCategory.module.css
kainino0x, line 1: The styling looks good in a local build for me. Could you update the screenshot (or remove it) since it's outdated right now?

src/pages/samples/[slug].tsx
```diff
@@ -13,55 +13,106 @@ type PageComponentType = {
...
+  };
+};
+
+export const pageCategories: PageCategory[] = [
+  createPageCategory('WebGPU Basics', webGPUBasicsPages),
+  createPageCategory('Feature Demos', featureDemoPages),
+  createPageCategory('Render Pass Demos', renderPassDemoPages),
+  createPageCategory('GPU Compute Demos', gpuComputeDemoPages),
+  createPageCategory('Web Platform Demos', webPlatformPages),
+  createPageCategory('Benchmarks', benchmarkPages),
```
Comment (kainino0x) [inline src/pages/samples/[slug].tsx:102]: I prefer the categorization I proposed (... of course, I proposed it). The reasoning behind it was:

- WebGPU Demos: The first thing people often want to see is complete demos, but at the same time they're less targeted as references because there's more code to look at. These are hard to categorize as render or compute because they all do a lot of different things in the same demo - it depends on what the reader is interested in finding. I also call them "demos" as in "complete demos of some things you can do with WebGPU vs "samples" which are more "reference code".
- WebGPU Basics: It's hard to subdivide this category because these are all pretty trivial samples using the WebGPU flavor of basic features of graphics APIs. Could be better titled "Basic Graphics" maybe.
- Important Graphics Features: These are particular API features that we want to highlight as being available in WebGPU. Render bundles in particular are somewhat special in WebGPU (a lot more important than native API equivalents). It's a bit subjective what goes here vs WebGPU Basics but I think these are the only two really important ones we have right now.
- Technique Implementations: this section really has nothing to do with WebGPU, except it's showing how some known technique can be reimplemented in WebGPU. For example all the interesting "cameras" code is in JS. So I made those separate. I sorted them by the problem domain, rather than whether they use RenderPipeline or ComputePipeline because many of them use compute to do graphics. (skinnedMesh would go in the Graphics Technique Implementations section.)

Comment (cmhhelgeson) [reply src/pages/samples/[slug].tsx:102]: I guess my contention for 'WebGPU Demos' would be that, although most of these samples are technically performing both render and compute operations, these samples seem to be tailored to demonstrate one over the other (i.e bitonicSort has a shader component, but is primarily meant to show how to perform a bitonic sort in compute. Same thing with boids, particles, and imageBlur. There is a rendered result of the compute operation, but the primary focus is on the compute operation). Without these distinctions, and by just lumping in such demos into a broad and difficult to parse category like 'WebGPU Demos', the need for categorization seems to disappear.

I do agree with you though that it would be good rename 'WebGPU Basics' to 'Graphics Basics' and that there should be a section that specifically addresses features unique to WebGPU (I've only really done WebGPU work so I wouldn't really know what those features are). I'll recategorize some of the samples to account for these new categories.

src/components/SampleCategory.module.css
cmhhelgeson: Will in a moment...

src/components/SampleCategory.module.css
cmhhelgeson: Categories aren't represented by an a element, this is just moving the styling of link elements from MainLayout.module.scss to SampleCategory, since, as of this commit, the style only applies to element's within SampleCategory's domain.

src/components/SampleCategory.module.css
cmhhelgeson: Noted, will remove.

src/pages/MainLayout.module.css
cmhhelgeson: Removed

src/components/SampleCategory.module.css
kainino0x: Oh, I see. It is still in MainLayout.module.scss though so it seems duplicated still.

src/pages/samples/[slug].tsx
Comment (kainino0x) [reply src/pages/samples/[slug].tsx:102]: The point about "WebGPU Demos" makes sense, I am fine with dissolving that category into other sections though I don't know how to do it. For example I believe A-buffer is compute heavy, but it's solving a graphics-specific problem.

I thought about having three categories but struggled to name them, but they probably divide cleanly if we can think of a way to title them:
- Non-graphics/GPGPU Techniques (bitonicSort, imageBlur?)
- Graphics techniques that don't use compute (IIRC cameras, shadowMapping, deferredRendering, cubemap, normalMap)
- Graphics techniques that do use compute (imageBlur?, A-buffer)

Posted what that would look like in https://github.com/webgpu/webgpu-samples/issues/357#issuecomment-1970157569

Comment (cmhhelgeson) [reply src/pages/samples/[slug].tsx:102]: I think this is a good categorization, only thing I would change is moving boids to GPGPU, since the visual component of that sample seems to be just providing a visual representation of an update that is mostly happening in compute. The sample would be practically identical if we were using canvas2D to render the boids instead of a shader (which, now that I mention it, could probably act as a good rationale for determining what samples belong in GPGPU and what samples belong in render+compute)

Comment (cmhhelgeson) [reply src/pages/samples/[slug].tsx:120]: Sorry this took so long, I've added what I think should be the final categorizations. I've also well added commented descriptions for each category to help future contributors delineate between these new categories and determine where their new samples may belong.

Comment (kainino0x) [reply src/pages/samples/[slug].tsx:122]: > the visual component of that sample seems to be just providing a visual representation of an update that is mostly happening in compute

I was thinking since it's a non-physical particle simulation, it doesn't really have any non-graphical application, so I feel it belongs in the "problem domain" with other graphics stuff. At the same time the compute part is very independent of the rendering (it's not part of the pipe of steps to put pixels on the screen, it could be run at a lower rate than rendering). So I am totally fine with putting it in GPGPU and I think your criteria for GPGPU draws a pretty nice line.

I was going to suggest documenting the categories before I read that you'd already done it. Great!

> Sorry this took so long

No rush!

And re everything else you said, I will finish reviewing tomorrow. :)

src/components/SampleCategory.module.css
```diff
@@ -0,0 +1,23 @@
+@keyframes openTriangle {
+  from { transform: rotate(0deg); }
+  to { transform: rotate(180deg); }
+}
+
+@keyframes closeTriangle {
+  from { transform: rotate(180deg); }
+  to { transform: rotate(0deg); }
+}
```
kainino0x, line 9: leftover

src/pages/samples/[slug].tsx
```diff
@@ -13,55 +13,126 @@ type PageComponentType = {
...
 };
 
+// Samples that demonstrate the GPGPU functionality of WebGPU. These samples generally
+// provide a visual representation of the result of a compute operation. Accordingly, the rendering
+// functionality of these samples exists soley to demonstrate the results of these operations.
+// As a general rule of thumb, samples within this category should be of two types:
+//   1. Visualizations of compute operations where the rendering is or could just as effectively be
+//      implemented within canvas2D with minimal changes to the code.
+//   2. Visualizations of compute features that are endemic to the domain of machine learning or A.I
+//      such as 'f16', convolution, backpropogation, etc.
```
Comment (kainino0x) [inline src/pages/samples/[slug].tsx:67]: - f16 is also very useful in graphics
- There are other non-graphics applications of compute like scientific computing, and these too can benefit from ML-focused extensions like dp4a and cooperative matrix multiply.
- It's likely we'll eventually have some samples with text or audio output instead of image output.

I think just point 1 is sufficient to define this category. How about:

```suggestion
// Samples that demonstrate the GPGPU functionality of WebGPU. These samples generally provide some
// user-facing representation (e.g. image, text, or audio) of the result of compute operations.
// Any rendering code is primarily for visualization, not key to the unique part of the sample;
// rendering could also be done using canvas2D without detracting from the sample's usefulness.
```

src/pages/samples/[slug].tsx
```diff
@@ -13,55 +13,126 @@ type PageComponentType = {
...
+  return {
+    title,
+    pages,
+    sampleNames: Object.keys(pages),
+  };
+};
+
+export const pageCategories: PageCategory[] = [
+  createPageCategory('Basic Graphics', graphicsBasicsPages),
+  createPageCategory('WebGPU Features', webGPUFeaturesPages),
```
Comment (kainino0x) [inline src/pages/samples/[slug].tsx:118]: Maybe "Feature Highlights"?

src/pages/samples/[slug].tsx
```diff
@@ -13,55 +13,126 @@ type PageComponentType = {
...
+  createPageCategory('GPGPU Demos', gpuComputeDemoPages),
+  createPageCategory('Graphics Techniques', graphicsDemoPages),
+  createPageCategory('Web Platform Demos', webPlatformPages),
```
Comment (kainino0x) [inline src/pages/samples/[slug].tsx:121]: Maybe "Web Platform Integration"?

src/pages/samples/[slug].tsx
```diff
@@ -13,55 +13,126 @@ type PageComponentType = {
...
   ),
-  imageBlur: dynamic(() => import('../../sample/imageBlur/main')),
-  shadowMapping: dynamic(() => import('../../sample/shadowMapping/main')),
   reversedZ: dynamic(() => import('../../sample/reversedZ/main')),
+  renderBundles: dynamic(() => import('../../sample/renderBundles/main')),
+};
+
+// A selection of samples demonstrating various graphics techniques, utilizing various features
+// of the WebGPU API, and often executing render and compute pipelines in tandem to achieve their
+// visual results.
```
Comment (kainino0x) [inline src/pages/samples/[slug].tsx:45]:
```suggestion
// visual results. The techniques demonstrated may even be independent of WebGPU (e.g. 'cameras').
```

src/pages/samples/[slug].tsx
Comment (kainino0x) [reply src/pages/samples/[slug].tsx:122]: I think all these categories look good.

kainino0x (approved):
LGTM % any open comments

src/pages/samples/[slug].tsx
Comment (cmhhelgeson) [reply src/pages/samples/[slug].tsx:45]: Added.

src/pages/samples/[slug].tsx
Comment (cmhhelgeson) [reply src/pages/samples/[slug].tsx:118]: I think WebGPU still needs to be specified. The natural question to ask with such a category would be "Feature Highlights of what?". I understand that using the word 'highlights' might help indicate that the category is not a comprehensive representation of all WebGPU specific features, but I think the word features doesn't do much to indicate the comprehesiveness of the samples in either direction.

src/pages/samples/[slug].tsx
Comment (cmhhelgeson) [reply src/pages/samples/[slug].tsx:121]: Yeah, this is a better name.

src/components/SampleCategory.module.css
cmhhelgeson: Removed

src/components/SampleCategory.module.css
cmhhelgeson: Removed from mainlayout.module.scss

cmhhelgeson:
This is good to go, just need somebody to merge it in.

src/pages/samples/[slug].tsx
Comment (kainino0x) [reply src/pages/samples/[slug].tsx:114]: OK, fine with me

PR #365, opened 2024-03-05
Fix redirection

beaufortfrancois:
Following https://github.com/webgpu/webgpu-samples/pull/364, this PR fixes redirection in production environment.
@austinEng Please review.

PR #367, opened 2024-03-05
Fix redirect for videoUploadingWebCodecs

austinEng:
Next.js redirects don't work for static-site generation because it can't generate a .html file for the page.

Follow-up on #364 and #365

### DIFF
--- a/next.config.js
+++ b/next.config.js
@@ -8,15 +8,6 @@ module.exports = {
   basePath: BASE_PATH,
   compress: true,
   reactStrictMode: true,
-  async redirects() {
-    return [
-      {
-        source: `${BASE_PATH}/samples/videoUploadingWebCodecs`,
-        destination: `${BASE_PATH}/samples/videoUploading?videoSource=videoFrame`,
-        permanent: true,
-      },
-    ]
-  },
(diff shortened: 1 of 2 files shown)


kainino0x (approved):

kainino0x:
@austinEng
https://webgpu.github.io/webgpu-samples/samples/videoUploadingWebCodecs
now redirects to

austinEng:
ugh yea.
somehow it still doesn't work.

Deployed gh page:
https://github.com/webgpu/webgpu-samples/blob/gh-pages/samples/videoUploadingWebCodecs.html

has
```
<meta http-equiv="refresh" content="0; url=/webgpu-samples/samples/videoUploading?videoSource=videoFrame"/>
```
in it

but yea, I get redirected to 
/samples/videoUploading?videoSource=videoFrame instead

austinEng:
OH. I see. in the sources panel, the redirect url is correct.

but if I break in DOMWindow.load - before redirect happens, it is wrong.

Basically, it seem JS is loading, changing the meta tag, and then it redirects... :-/

beaufortfrancois:
Thank you @austinEng and @kainino0x for addressing this issue!

PR #368, opened 2024-03-06
Attempt to fix videoUploadingWebCodecs redirect again

austinEng:
define BASE_PATH so we don't try to evaluate it at runtime. It is only defined in the environment at build time.

follow-up to #367


kainino0x (approved):
seems plausible

kainino0x:
works now! once you get past the browser cache

PR #369, opened 2024-03-06
Use iframes

greggman:
This PR removes react and next. It allows a sample to be pure JavaScript (though there are none at the moment, all the samples are in typescript). It allows samples to be standalone so that they run separate from the sample selection menu. (which is how they are able to run in iframes).

Note: The last 4 PRs are not merged (they're overwritten). The 3 related to redirects, AFAICT, are irrelevant after this PR. The group descriptions I can add back but this is taking much longer than expect (Have spent ~30hrs) so crossing my fingers this is ok. Open to suggestions.

Note: Originally I had changed all of the `xxx.wgsl` files to `xxx.wgsl.js` and surrounded the shaders with 

```
export default `
   ...wgsl...
`;
```

but I felt there were be resistance so I left it as is. We can revisit that. My issue with it is is it requires a build step and special rules. Of course typescript requires a build step too but if switch the `.wgsl` to `.wgsl.js` then all you need is typescript to build and run the examples. You don't need some bundler to hack the `.wgsl` files into includable files. That seems more useful for people wanting to use the samples elsewhere? OTOH the plugin to make it work is small in most build systems so I'm not 100 sure what I feel about that ATM.


greggman:
there's a live version here: https://greggman.github.io/webgpu-samples-iframes/

beaufortfrancois:
@greggman Can you make sure https://webgpu.github.io/webgpu-samples/samples/videoUploadingWebCodecs still redirects to https://webgpu.github.io/webgpu-samples/samples/videoUploading?videoSource=videoFrame?
See https://github.com/webgpu/webgpu-samples/pull/367

public/css/MainLayout.css
```diff
@@ -48,40 +49,37 @@
...
 
   .panel .panelContents {
     display: block;
     transition: max-height 0.3s ease-out;
-    max-height: 0px;
-  }
-
-  .panel[data-expanded='false'] {
-    overflow: hidden;
+    gmax-height: 100%;
```
austinEng, line 74: I couldn't find anything that tells me what gmax-height is. Is this valid?

sample/animometer/index.html
```diff
@@ -0,0 +1,27 @@
+<!DOCTYPE html>
```
austinEng, line 1: So previously https://webgpu.github.io/webgpu-samples/samples/animometer showed the whole rest of the page, but now it'll only show the iframe.

This probably isn't ideal since all existing links will go to a page that could be quite different from what it used to. It's.. annoying but could we perhaps make all the existing pages redirect to https://webgpu.github.io/webgpu-samples/?sample=animometer

The individual sample pages could maybe be at sample/animometer/page.html - it's probably ok to have the .html suffix, I think.

Finally, it's not discoverable that the standalone page exists. Perhaps the iframe version and the standalone versions should link to each other - that can be done later though

sample/videoUploadingWebCodecs/index.html
```diff
@@ -0,0 +1,9 @@
+<!DOCTYPE html>
+<html>
+  <head>
+      <meta
+        http-equiv="refresh"
+        content="0;URL='../../?sample=videoUploading?videoSource=videoFrame'"
```
austinEng, line 6: TIL you can do two ? for queries. Seems fine since it works, but I would normally expect `../../?sample=videoUploading&videoSource=videoFrame`

sample/animometer/index.html
austinEng: oh - actually I've realized the old path is https://webgpu.github.io/webgpu-samples/samples/animometer and the new standalone path is https://webgpu.github.io/webgpu-samples/sample/animometer 

so maybe we can have:
https://webgpu.github.io/webgpu-samples/samples/animometer redirect to https://webgpu.github.io/webgpu-samples/?sample=animometer

and https://webgpu.github.io/webgpu-samples/sample/animometer be the iframe

austinEng:
Navigation bar seems cut off on mobile when expanded
greggman github io_webgpu-samples-iframes__sample=helloTriangle(Pixel 7)

On the home page, the background doesn't extend all the way
greggman github io_webgpu-samples-iframes_(Pixel 7)

public/css/MainLayout.css
greggman: fixed (that's how I comment out styles. Forgot to remove)

sample/animometer/index.html
greggman: done

greggman:
fixed screenshots

sample/videoUploadingWebCodecs/index.html
greggman: One is query for the framework (which sample to show), other is a query for the iframe's page. They aren't related.

kainino0x: It parses as `sample` = `videoUploading?videoSource=videoFrame` (in URLSearchParams).

kainino0x:
> My issue with it is is it requires a build step and special rules.

How about loading the shaders using fetch instead?

index.html
```diff
@@ -0,0 +1,86 @@
...
+          <hr>
+          <div id="samplelist"></div>
+          <hr>
+          <h3 style="margin-bottom: 5px">Other Pages</h3>
+          <ul class="exampleList">
+            <li>
+              <a
+                rel="noreferrer"
+                target="_blank"
+                href="/workload-simulator.html"
```
austinEng, line 51: I think this link should be:

```suggestion
                href="./workload-simulator.html"
```

austinEng (approved):
this is a huge diff and I tried to look through most of the non-sample parts but may have missed things. I skimmed most of the existing webgpu code.

I think we may as well land it and follow-up on any improvements / changes since the high level structure probably won't change significantly.

index.html
greggman: done

Issue #374, opened 2024-03-07
Missing redirect


beaufortfrancois:
As raised in https://github.com/webgpu/webgpu-samples/issues/369#issuecomment-1980247086, it would be nice if we could make sure https://webgpu.github.io/webgpu-samples/samples/videoUploadingWebCodecs still redirects to https://webgpu.github.io/webgpu-samples/samples/videoUploading?videoSource=videoFrame

Context: https://github.com/webgpu/webgpu-samples/pull/367

@greggman

beaufortfrancois:
We can close this issue since https://github.com/webgpu/webgpu-samples/pull/376 addressed it.

PR #376, opened 2024-03-07
Fix redirect for samples/videoUploadingWebCodecs

greggman:
I fixed this once but then a later change broke it 😢

#374


greggman:
ps: the rollup stuff is just lint-fix stuff. I should add it to the lint rule.

beaufortfrancois:
Thank you!

Issue #362, opened 2024-03-05
Link to or embed other known samples


kainino0x:
- @toji
  - https://spookyball.com/
  - https://toji.github.io/webgpu-metaballs/
  - https://toji.github.io/webgpu-clustered-shading/
  - https://toji.github.io/pristine-grid-webgpu/
  - https://toji.github.io/webgpu-bundle-culling/
- I know there are others but not off the top of my head

It may become possible to embed these once #355 is done (though idk if github.io allows COEP)

greggman:
Put up a PR. It works. Thoughts?

It's not clear to me if linking to external examples a good thing or not. I think Brandon's examples are great but I once you set the example of linking to external examples then everyone will want their sample included. Maybe that's a good thing though?

Note: the PR also tries to grab the source which works for spookyball but there's many files (didn't list them all. Just wanted to see if it worked). We could not show any source, or show something else (link to article/github).

@kainino0x @toji @beaufortfrancois @Kangz @austinEng

kainino0x:
> It's not clear to me if linking to external examples a good thing or not. I think Brandon's examples are great but I once you set the example of linking to external examples then everyone will want their sample included. Maybe that's a good thing though?

I was thinking that's probably a good thing. Right now we have people wanting to add samples to this repo which means we have to take on some maintenance burden. Easier if we can link to them. The number of links will probably get pretty big though, so maybe external samples would be a separate section in the sidebar.

That said, it occurs to me that linked samples could change underneath us - disappear, break, or even get replaced with malicious content.
A middle ground could be to accept samples as Git submodules so we're pinned to a revision?

greggman:
I think the samples changing is not a big deal. People will file an issue and we can fix or remove.

I don't think we don't want to rehost them (which might include having to build and maintain their build infra). I suppose you meant to pin their live repo (eg gh-pages) as a submodule but that would mean having to review updates often if they're iterating on their sample.

toji:
I agree that we don't want to re-host any content. It does open us up for some minor unexpected breakage, but the maintenance burden otherwise is going to be pretty high and driven by external forces.

greggman:
done?

PR #378, opened 2024-03-07
make external links work

greggman:
This is a work in progress so no review needed yet. Though of course feedback welcome. Of course a different category would be good. I just wanted to see it work so stuck it in. I can add the rest if we've decided to go forward with this.

It's not clear to me if linking to external examples a good thing or not. I think Brandon's examples are great but I once you set the example of linking to external examples then everyone will want their sample included. Maybe that's a good thing though?

#362

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -91,7 +91,7 @@ function setSourceTabHash(event: PointerEvent, sourceInfo: SourceInfo) {
 
 // Non authoritative test that url is for same domain
 function isSameDomain(url: string) {
-  return !/^\w+:\/\//.test(url);
+  return new URL(url, window.location.href).origin === window.location.origin;
 }
 
 // That current sample so we don't reload an iframe if the user picks the same sample.
--- a/src/main.ts
+++ b/src/main.ts
@@ -5,6 +5,11 @@ import { EditorView } from '@codemirror/view';
 import { EditorState } from '@codemirror/state';
 import { javascript } from '@codemirror/lang-javascript';
 import { basicSetup } from 'codemirror';
+import { Converter } from 'showdown';
+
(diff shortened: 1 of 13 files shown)


src/main.ts
```diff
@@ -84,6 +84,11 @@ function setSourceTabHash(event: PointerEvent, sourceInfo: SourceInfo) {
   setSourceTab(sourceInfo);
 }
 
+// Non authoritative test that url is for same domain
+function isSameDomain(url: string) {
```
kainino0x, line 88: nit: I think `new URL(url, window.location).origin === window.location.origin` should work

kainino0x: Alternate nit: rename isSameDomain to isRelativeURL or similar

greggman:
okay, updated with all 5 examples. Please review.

src/main.ts
greggman: choose the first.

greggman:
I agree. I was trying to think of some but I don't know any off the top of my head

I could put a few from webgpufundamentals but I could just copy them here. One that comes to mind is the multiple canvas example. It could use better models on the sample site whereas on the fundamentals site I wanted to keep it simple.

https://webgpufundamentals.org/webgpu/webgpu-multiple-canvases-x200-optimized.html

Another is points but again, can just copy it

https://webgpufundamentals.org/webgpu/webgpu-points-w-size.html

And maybe give it some GUI options for the rest

https://webgpufundamentals.org/webgpu/webgpu-points-w-rotation.html

https://webgpufundamentals.org/webgpu/webgpu-points-3d.html

https://webgpufundamentals.org/webgpu/webgpu-points-3d-fixed-size.html

PR #380, opened 2024-03-08
Style adjustments

cmhhelgeson:
Just fixed scrollbar and tried centering data-tooltip popup. Also made descriptions more curt.


greggman:
It seems there's still an unrelated image file in this PR?

If I might make some suggestions, there are many different workflows in git.

The one I use is this

```
git remote add upstream https://github.com/webgpu/webgpu-samples.git
```

Then, at anytime I can do this

```
git fetch upstream main     # gets the latest version and stores it in the .git folder
git rebase upstream/main    # merges whatever changes I have locally on top of the current version
```

Following that path we get rid of all of this

Screenshot 2024-03-07 at 22 12 30

Also, before you upload you could do `npm run lint` and it would show you if there are any errors that you can fix locally before uploading. `npm run fix` will try to auto-fix many of those issues. Commit everything before you do `npm run fix` just as a precaution.

Also, It's not a big deal but we get interrupted with notifications every time new updates are pushed so fixing things locally and then updating would mean less notifications for others.

You could even do this (I do)

```
echo > .git/hooks/pre-push "npm run lint"
chmod u+x .git/hooks/pre-push   # only on mac/linux
```

This will run lint before pushing and stop if there are any errors.

cmhhelgeson:
> It seems there's still an unrelated image file in this PR?
> 
> If I might make some suggestions, there are many different workflows in git.
> 
> The one I use is this
> 
> ```
> git remote add upstream https://github.com/webgpu/webgpu-samples.git
> ```
> 
> Then, at anytime I can do this
> 
> ```
> git fetch upstream main     # gets the latest version and stores it in the .git folder
> git rebase upstream/main    # merges whatever changes I have locally on top of the current version
> ```
> 
> Following that path we get rid of all of this
> 
> Screenshot 2024-03-07 at 22 12 30
> Also, before you upload you could do `npm run lint` and it would show you if there are any errors that you can fix locally before uploading. `npm run fix` will try to auto-fix many of those issues. Commit everything before you do `npm run fix` just as a precaution.
> 
> Also, It's not a big deal but we get interrupted with notifications every time new updates are pushed so fixing things locally and then updating would mean less notifications for others.
> 
> You could even do this (I do)
> 
> ```
> echo > .git/hooks/pre-push "npm run lint"
> chmod u+x .git/hooks/pre-push   # only on mac/linux
> ```
> 
> This will run lint before pushing and stop if there are any errors.

Apologies, I will try to be more conscientious in the future.

PR #334, opened 2023-11-27
Remove entryPoint if not required

beaufortfrancois:
See https://github.com/gpuweb/gpuweb/pull/4387/

⚠️ Do not merge this PR until Chrome 121 reaches stable channel.

PR #382, opened 2024-03-08
Add an MSDF-based text rendering sample

toji:
This sample has been languishing in a branch for a while because react's handling of asset files prevented it from working correctly. Now that react has been removed from the sample pages, though, it works as expected! Thanks @greggman!

Updated to match the newer sample format. Please let me know if I missed anything, but it looks like everything is working as expected when I run it locally.


sample/textRenderingMsdf/main.ts
```diff
@@ -0,0 +1,310 @@
...
+          },
+        ],
+      },
+    ],
+  },
+  fragment: {
+    module: device.createShaderModule({
+      code: vertexPositionColorWGSL,
+    }),
+    entryPoint: 'main',
```
greggman, line 149: same as above

sample/textRenderingMsdf/main.ts
```diff
@@ -0,0 +1,310 @@
...
+new Float32Array(verticesBuffer.getMappedRange()).set(cubeVertexArray);
+verticesBuffer.unmap();
+
+const pipeline = device.createRenderPipeline({
+  layout: 'auto',
+  vertex: {
+    module: device.createShaderModule({
+      code: basicVertWGSL,
+    }),
+    entryPoint: 'main',
```
greggman, line 124: I think in general we're not longer using `entryPoint`? https://github.com/webgpu/webgpu-samples/pull/334

up to you though

sample/textRenderingMsdf/main.ts
```diff
@@ -0,0 +1,310 @@
...
+    targets: [
+      {
+        format: presentationFormat,
+      },
+    ],
+  },
+  primitive: {
+    topology: 'triangle-list',
```
greggman, line 157: remove? (it's the default).  Or not, up to you

sample/textRenderingMsdf/main.ts
```diff
@@ -0,0 +1,310 @@
...
+    },
+  ],
+});
+
+const renderPassDescriptor: GPURenderPassDescriptor = {
+  colorAttachments: [
+    {
+      view: undefined, // Assigned later
+
+      clearValue: { r: 0.0, g: 0.0, b: 0.0, a: 1.0 },
```
greggman, line 203: question: Do you think we should prefer `[0, 0, 0, 1]` do you prefer this longer format?  My impression is maybe this longer format was the only format originally?

sample/textRenderingMsdf/meta.ts
```diff
@@ -0,0 +1,19 @@
+export default {
+  name: 'Text Rendering - MSDF',
+  description: `This example uses multichannel signed distance fields (MSDF) to render text. MSDF
+      fonts are more complex to implement than using Canvas 2D to generate text, but the resulting
+      text looks smoother while using less memory than the Canvas 2D approach, especially at high
+      zoom levels. They can be used to render larger amounts of text efficiently.
+
+      The font texture is generated using <a href="https://msdf-bmfont.donmccurdy.com/">Don
```
greggman, line 8: Up to you but this is markdown so maybe markdown format `[Don McCurdy's MSDF font generation tool](https://msdf-bmfont.donmccurdy.com/)`.  I only suggest that because I don't want to encourage arbitrary HTML.

sample/textRenderingMsdf/msdfText.wgsl
```diff
@@ -0,0 +1,81 @@
+// Positions for simple quad geometry
+const pos : array<vec2f, 4> = array<vec2f, 4>(
```
greggman, line 2: do you want to use the shorter format?

```
const pos = array( vec2f(0, -1), vec2f(1, -1), vec2f(0, 0), vec2f(1, 0));
```

greggman (approved):

sample/textRenderingMsdf/main.ts
greggman: lol. I thought because this was at line 146 and the comment below at line 121 that the one below would show up first but apparently github orders by the comment entry order 🤷‍♂️

greggman:
oh, and you might want to `npm run fix`

sample/textRenderingMsdf/main.ts
toji: A lot of this was just copied from an earlier version of the rotating cube sample, but I'm happy to update it as I go! My preference has always been for the more concise array format if I'm typing it directly. I've hit a couple of places where it's made more sense to use the longer dictionary versions, but that's almost always when processing some sort of externally provided values that are already named.

sample/textRenderingMsdf/msdfText.wgsl
toji: Yup.

sample/textRenderingMsdf/meta.ts
toji: Updated to use markdown, though it threw me off for a second because I had the text indented in the backticks, so it read them as a literal. The indented text was then treated as a code block, and didn't format the link correctly. Easy to fix once I realized what was going on.

toji:
Thanks for the feedback! Updated everything, rebased, and cleaned up a few additional TypeScript complaints as well.

sample/textRenderingMsdf/meta.ts
```diff
@@ -0,0 +1,18 @@
+export default {
+  name: 'Text Rendering - MSDF',
+  description: `This example uses multichannel signed distance fields (MSDF) to render text. MSDF
```
greggman, line 3: I'm guessing you already know this but just in case... you can do this if you want

```js
{
  description: `\
This example uses multichannel signed distance fields (MSDF) to render text. MSDF
fonts are more complex to implement than using Canvas 2D to generate text, but the resulting
text looks smoother while using less memory than the Canvas 2D approach, especially at high
zoom levels. They can be used to render larger amounts of text efficiently.`,
}
```

not suggesting you change anything. Just passing it on.

kainino0x: I think we should trim the description if we don't already, then you don't have to worry about the backslash. It also allows putting the closing `` `, `` on a separate line. This is what we do in test descriptions in CTS.

greggman: If trim means just `description.trim()` I'm for it but it wouldn't have helped the original issue here which was a description like this.

```
{
  description: `
     This is the description
     bla bla bla bla bla bla
     bla bla bla bla bla bla
     bla bla bla bla bla bla
  `
}
```

Because that leading space is sensitive in markdown. We could try to trim every line the same amount that feels like overkill and likely to break. I think just knowing the text is markdown and knowing markdown is space sensitive is enough. Maybe we should rename `description` to `descriptionMarkdown`.  A simple trim just for leading and training space though is unlikely to break anything. It's unlikely the first line would be indented intentionally.

PR #386, opened 2024-03-08
Make the workload-simulator work as a sample

greggman:
Before it was specially hand coded. Now it can exist as a sample like other samples.

It wasn't clear to me if it should open in a iframe or pop out. I made it pop out since that is what it used to do.

Also, I could put a re-direct at the root if you think one is needed so links to the old one are still valid though I think people will figure it out without and it seems like clutter. wdyt?


toji (approved):
LGTM! Nice that this is normalized with the rest of the samples now.

PR #391, opened 2024-03-11
Remove React guard from worker sample

toji:
Removes some code from the worker sample that only existed to catch errors when the code was executed twice by React. Now that React has been removed it's no longer necessary.

~~Also removed the white background style for sample iframes, which looked strange and I assume to be a mistake? If this was intentional let me know and I'll revert this bit.~~

### DIFF
--- a/sample/worker/main.ts
+++ b/sample/worker/main.ts
@@ -17,25 +17,17 @@ worker.addEventListener('message', (ev) => {
   }
 });
 
-try {
-  // In order for the worker to display anything on the page, an OffscreenCanvas must be used.
-  // Here we can create one from our normal canvas by calling transferControlToOffscreen().
-  // Anything drawn to the OffscreenCanvas that call returns will automatically be displayed on
-  // the source canvas on the page.
-  const offscreenCanvas = canvas.transferControlToOffscreen();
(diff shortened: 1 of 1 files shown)


public/css/SampleLayout.css
```diff
@@ -13,7 +13,7 @@
   height: 100%;
   border: none;
   display: block;
-  background-color: #fff;
+  /*background-color: #fff;*/
 }
```
greggman, line 17: I don't think this should be removed. iframes (this seems like a bug in the spec to me) get their background color from their parent which is bad. The page itself might do have the default black text in which case you'd get black on black text. This is especially true for external samples. They all expect the default is white.

I think rather, add 

```css
:root {
  color-scheme: light dark;
}
```

to all the `sample/**/index.html` files? Let the samples be responsible for what background color they want.

Also, I think we should look into making the menus be responsive to dark/light prefs in some future PR.

wdyt?

greggman (approved):

public/css/SampleLayout.css
toji: Agreed. I don't think the white backgrounds look great, but we can have the samples decide that for more consistency.

Issue #392, opened 2024-03-11
Add light/dark support to samples and menus


greggman:
The menus are currently always dark on light

<img width="284" alt="Screenshot 2024-03-11 at 13 45 15" src="https://github.com/webgpu/webgpu-samples/assets/234804/c889a18f-e605-4be6-88bc-cfae05467409">

They should probably be responsive to the user's preferences

1. Put

   ```css
   :root {
     color-scheme: light dark;
   }
   ```

   in `styles.css` and in all the `sample/**/index.html` files. This makes the browser choose light are dark colors based on the user's preferences for not just the background color but also for scrollbars, `<input type="???">` etc...

2. add variables for colors in the `styles.css` and use them where appropriate

   ```css
   :root {
     /* these are the light colors */
     --some-color: #fe803f;
     --some-other-color: #fe803f;
   }
   @media (prefers-color-scheme: dark) {
     /* these are the dark colors */
     :root {
       --some-color: #2e401f;
       --some-other-color: #2e401f;
     }
   }
   ```

   then use the variables where appropriate

   ```css
   [data-tooltip]::after {
     background-color: var(--some-color);
     ...
   ```

   Less colors = better

3. The `<iframe>` must have a white background
 
   Pages default to white. Pages in the iframe may be external and may default to black text so we can't se the `<iframe>`'s color to anything other than white else it may be incompatible with the sample. For the samples in this repo, see (1) above.

PR #395, opened 2024-03-12
support light/dark

greggman:
No worries if this is not desired. I just thought I'd take a stab at it.

Here's the results

Screenshot 2024-03-11 at 21 19 23

---

Screenshot 2024-03-11 at 21 21 33

I can look into the codemirror stuff later.

#392

note: read this and 😭
https://fvsch.com/transparent-iframes

This works for Safari and Chrome but AFAICT there is no solution on Firefox. There, samples are always white ATM. Maybe there is though and I just didn't find it.

### DIFF
--- a/index.html
+++ b/index.html
@@ -62,7 +62,7 @@
           <div class="sampleContainer"></div>
         </div>
         <nav id="code" class="sourceFileNav">
-          <div>
+          <div class="sourceFileScrollContainer">
             <ul id="codeTabs"></ul>
           </div>
         </nav>
--- a/public/css/HomePage.css
+++ b/public/css/HomePage.css
@@ -1,4 +1,41 @@
+:root {
+  color-scheme: light dark;
+
+  /* these are the light colors */
+  --background: #eee;
+
+  --code-background: #ccc;
+  --link: #045b88;
+
+  --panel-color: #000;
+  --panel-background: #ddd;
+  --panel-emphasis: rgb(20, 62, 84);
+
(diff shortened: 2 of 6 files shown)


greggman:
note: I pulled out the left/right scrolling source tabs. It's a nice idea but

1. it's not discoverable. You have to scroll left and right with a trackpad/mouse. Clicking/hovering never shows anything so you don't know it can do it.

2. it's not mobile friendly. AFAICT You can only scroll left/right once the scrollbar appears but there is no way to make it appear on mobile

At least that's how it seemed.  It's fine if it's put back in but I didn't have time to address the issues. One issue might be to add add `<` and `>` buttons. Another might a drop down when the list is too large (devtools does this with `»` menu). Or we can just leave it as is. The list now shows just `filename` and no path which means the tabs are shorter so more often than before they all fit.

toji:
This all looks good to me, and I like that the whole site is responsive to theme preferences now. The source tab behavior seems OK to me for now, in that it simply starts wrapping the tabs. Not particularly elegant, but functional.

Screenshot 2024-03-12 10 57 14 AM

The one issue I found, which can be addressed in a follow up if you'd prefer, is that when the side bar collapses on a narrow window or on mobile the hamburger icon is not responsive to the theme, It looks OK in the light theme but is so dark that you can hardly see it in the dark theme.

Screenshot 2024-03-12 10 54 58 AM (1)

toji (approved):

Issue #396, opened 2024-03-12
Re-add scrolling tabs


greggman:
I removed these temporarily because they are not discoverable and not mobile friendly (AFAICT) but I'm in the process of adding them back.

greggman:
work in progress:

https://github.com/greggman/webgpu-samples/pull/new/scroll-tabs

still needs styling

https://github.com/webgpu/webgpu-samples/assets/234804/ba5fb952-3e69-4825-8bad-a0d297a98ae3

PR #403, opened 2024-03-14
add back in scrolling source tabs

greggman:
I put `<` and `>` on the left and right. They should work on touch screens. Also made it so if you click a tab that is partially off screen it will get put on screen. Also if the url has a `#<filename>` it will start with that tab on screen.

https://github.com/webgpu/webgpu-samples/assets/234804/abc489b6-d2fe-4acc-8634-473eed16057b

### DIFF
--- a/index.html
+++ b/index.html
@@ -62,9 +62,11 @@
           <div class="sampleContainer"></div>
         </div>
         <nav id="code" class="sourceFileNav">
-          <div class="sourceFileScrollContainer">
+          <div class="sourceLR" id="sourceL">&lt;</div>
+          <div id="sourceTabs" class="sourceFileScrollContainer">
             <ul id="codeTabs"></ul>
           </div>
+          <div class="sourceLR" id="sourceR">&gt;</div>
         </nav>
         <div id="sources"></div>
       </main>
--- a/public/css/SampleLayout.css
+++ b/public/css/SampleLayout.css
@@ -47,12 +47,18 @@
   transform: translateY(0.25em);
 }
 
+nav.sourceFileNav {
+  display: flex;
+  align-items: flex-start;
+}
+
 nav.sourceFileNav ul {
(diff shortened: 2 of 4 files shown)


greggman:
#396

greggman:
It's been driving me crazy but sometimes the scrollbar is under the tabs, and sometimes it's at the bottom but overlaid on top of the tabs. Compare this video to the one above

https://github.com/webgpu/webgpu-samples/assets/234804/85caef3b-44e4-4ad6-bcca-c905355fbe90

I don't know what's causing the difference. This video is from syncing to the repo on my home computer. It's also the reason I originally cancelled the PR this morning. It was working at home (scrollbar overlaid on tabs) but then at work they were under/below. I reset back to before I added iframes and they were under/below so made a few tweaks to make the side buttons align and put up the PR.

Now I get home, sync, and the scrollbars are back to overlaid. I wonder if it's a OS setting (i'll have to check at work tomorrow).

Screenshot 2024-03-13 at 21 34 12

In any case, it should be fine to approve I think. Unless you have any other issues you want addressed.

greggman:
I occured to me a few minutes after the last comment that I could just set that setting on my home machine. I verified that's the difference. If I set "show scroll bars: always" then the bars appear under the tabs. If set to "automatically based on mouse or trackpad" then they are drawn overlaid on the tabs

toji (approved):
LGTM.

I don't know what could reasonably be done about the differing scroll bar behavior. I guess either the navigation buttons on either side of the tabs could be extended down to no longer leave a gap between them and the content or the scrollbar could be extended under the buttons, but it would have to do so in a way that while it encompassed them it didn't actually scroll them. Both sound unnecessarily tricky, and probably platform dependent.

In any case, I don't think that it's problematic as-is, we're really talking about a minor aesthetic quirk.

PR #414, opened 2024-04-03
Use a simpler `clearValue`

greggman:
This is obviously a personal preference but, ... I feel like the `{ r: ?, g: ?, b: ?, a: ? }` format is left over from before the time when an array of numbers was allowed?

The array of numbers is arguably more useful since all kinds of things can easily provide an array of numbers. So, it seems like it would be good for the samples to show the more useful usage. That way devs are not mislead into thinking they need to take some array of numbers and convert it into the more verbose format.

wdyt?

### DIFF
--- a/sample/a-buffer/main.ts
+++ b/sample/a-buffer/main.ts
@@ -122,7 +122,7 @@ const opaquePassDescriptor: GPURenderPassDescriptor = {
   colorAttachments: [
     {
       view: undefined,
-      clearValue: { r: 0, g: 0, b: 0, a: 1.0 },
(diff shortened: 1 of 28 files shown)


austinEng (approved):
I agree this is cleaner

kainino0x (approved):
Personally I prefer using the [0,0,0,0] syntax always. I actually tried to remove the alternate syntax for everything other than colors before 1.0, but there wasn't appetite to do it. I'd probably have been happy to remove it for colors too, but it was not as easy of a sell because for BGRA formats it's still [r,g,b,a] and that's can be a tiny bit confusing.

Issue #150, opened 2021-11-25
how to use StencilTest


hpugis:
A simple demo can be provided？
I tried it, but it didn't work
``

hpugis:
a simple demo 
`
async function init() {
  const gpu = navigator.gpu; 
  const adapter = await gpu.requestAdapter();
  const device = await adapter.requestDevice();
  
  const cvs = document.createElement('canvas');
  cvs.width = 1024;
  cvs.height = 768;
  document.body.appendChild(cvs);
  const depthTexture = device.createTexture({
    size: {
      width: cvs.width,
      height: cvs.height,
      depthOrArrayLayers: 1
    },
    format: "depth24plus-stencil8",
    usage: GPUTextureUsage.RENDER_ATTACHMENT
  });
  const ctx = cvs.getContext('webgpu');
  const swapChainFormat = "bgra8unorm";
  const presentationFormat = ctx.getPreferredFormat(adapter);
  ctx.configure({
    device: device,
    format: swapChainFormat
  });
  let data=new Float32Array(
    [
      -1.0, -1.0, 0.0, 1.0,
      0.0, 1.0, 0.0, 1.0,
      1.0, -1.0, 0.0, 1.0
    ]);
  let vertexBuffer = device.createBuffer({
    size: data.byteLength,
    usage: GPUBufferUsage.VERTEX | GPUBufferUsage.COPY_DST
  });
  device.queue.writeBuffer(vertexBuffer, 0, data);
  ///////////////////////////////////////////////////////////////////////////////////////////////////////////////////
  const uniformsBindGroupLayout = device.createBindGroupLayout({
    entries: [
      {
        binding: 0,
        visibility: GPUShaderStage.VERTEX,

        buffer: {
          type: 'uniform',
        },
      }
    ]
  });
  const uniformsBindGroupLayout1 = device.createBindGroupLayout({
    entries: [
      {
        binding: 0,
        visibility: GPUShaderStage.VERTEX,

        buffer: {
          type: 'uniform',
        },
      }
    ]
  });
  const matrixSize = 16 * Float32Array.BYTES_PER_ELEMENT; // 4x4 matrix
  const uniformBufferSize = matrixSize ;

  const uniformBuffer = device.createBuffer({
    size: uniformBufferSize,
    usage: GPUBufferUsage.UNIFORM | GPUBufferUsage.COPY_DST,
  });
  const uniformBuffer1 = device.createBuffer({
    size: uniformBufferSize,
[...]

PR #415, opened 2024-04-05
Stencil Mask Example

cmhhelgeson:
EDITED 4/3/24

A simple masking example leveraging the stencil buffer. Masks can be positioned with the mouse, resized with the scroll wheel, and inverted.

The example is intended to demonstrate how users can modify existing code with minimal adjustments to create masks with the stencil, render into different scenes, and leverage the capabilities of the render pipeline to directly modify the stencil buffer across one or more passes by disabling the pipeline's color and depth write functionality.
Capture

### DIFF
--- a/sample/stencilMask/index.html
+++ b/sample/stencilMask/index.html
@@ -1,0 +1,30 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta charset="utf-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
+    <title>webgpu-samples: stencilMask</title>
+    <style>
+      :root {
+        color-scheme: light dark;
+      }
+      html, body {
+        margin: 0;      /* remove default margin */
+        height: 100%;   /* make body fill the browser window */
+        display: flex;
+        place-content: center center;
+      }
+      canvas {
+        width: 600px;
+        height: 600px;
+        max-width: 100%;
+        display: block;
+      }
+    </style>
+    <script defer src="main.js" type="module"></script>
+    <script defer type="module" src="../../js/iframe-helper.js"></script>
+  </head>
+  <body>
+    <canvas></canvas>
+  </body>
+</html>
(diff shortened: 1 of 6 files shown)


cmhhelgeson:
Linking to relevant issue #150. Perhaps the reflective floor demo could also be added.

cmhhelgeson:
Hi, I just wanted to check in and see if there was anything more I could do here, of if a different kind of sample would be preferred (reflective floor, orbit around a portal, etc.)

greggman:
Sorry, I looked at this a month ago and it seemed not finished so I assumed you were planning to post more? The code references `fullscreenQuad.wgsl` but no such file exists. It also references `positionQuad.vertex.wgsl` but no such file exists. It also claims to move the mask with the mouse but that doesn't seem to happen.

Personally, I'm not sure it's a good example of what to use a stencil for. Blending between two images is more generally done with alpha channels since you can choose the level of blending per pixel. Rendering to the stencil with an SDF also seems unusual. Of course just seeing a stencil mask work is a probably a good example. I was just thinking it would be nice to see a practical example vs a contrived example. Unfortunately I couldn't think of one.

cmhhelgeson:
> Sorry, I looked at this a month ago and it seemed not finished so I assumed you were planning to post more? The code references `fullscreenQuad.wgsl` but no such file exists. It also references `positionQuad.vert.wgsl` but no such file exists. It also claims to move the mask with the mouse but that doesn't seem to happen.
> 
> Personally, I'm not sure it's a good example of what to use a stencil for. Blending between two images is more generally done with alpha channels since you can choose the level of blending per pixel. Rendering to the stencil with an SDF also seems unusual. Of course just seeing a stencil mask work is a probably a good example. I was just thinking it would be nice to see a practical example vs a contrived example. Unfortunately I couldn't think of one.
> 
> At the very least, what do you think about making the 2 scenes more varied? As it is, 4x4 grid of blue cubes vs a 4x4 rainbow cubes makes it seems like the effect is changing the cubes vs masking one rendering vs another. In other words, maybe one scene is 4x4 cubes and the other scene is a rotating spiral of spheres. (or something?) just throwing out ideas. [for example](https://greggman.github.io/webgpu-utils/examples/stencil-cube.html)
> 
> Another idea would be to look at some common uses for the canvas 2d mask

Wouldn't it just make more sense to port that existing sample over to webgpu-samples?

PR #418, opened 2024-05-20
fix wrong camera initilization direction

LPEdwin:
recalculateAngles has to be called with the back direction, but was falsely called with the forward direction in the constructor


ben-clayton (approved):
Good catch!

PR #419, opened 2024-05-25
Fix orientation of cubemap sample, add clearer attribution

### LINKED ISSUE #341: Is the cubemap example correct? (Flipped church)
Thanks for the cubemap example!

But I am a bit confused about the position of the church. When I arrange the cubemap images in the standard "+" format, I get the first image below. You can see in the -Z tile that the church is to the **left** of the bridge. Now look at the second image, which shows your example. The church is on the **right** of the bridge. Is the mapping correct? Currently the WebGPU documentation does not document the "cube" texture convention in detail. Does it match the OpenGL cubemap conventions in the link below? Thanks for any clarification!

https://www.khronos.org/opengl/wiki/Cubemap_Texture

bridge

Screenshot 2023-12-20 at 9 55 22 PM

kainino0x:
Fixes #341

### DIFF
--- a/public/assets/img/cubemap/credit.txt
+++ b/public/assets/img/cubemap/credit.txt
@@ -1,0 +1,1 @@
+Cubemap image available under a Creative Commons Attribution 3.0 Unported License at https://www.humus.name/index.php?page=Textures&ID=58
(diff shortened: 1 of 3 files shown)


greggman (approved):
lgtm

PR #422, opened 2024-06-03
feat: add entryPoints to fix err

Djjanks:
Add entrypoints to GPURenderPipelineDescriptor in modules to fix error on https://webgpu.github.io/webgpu-samples.

image

### DIFF
--- a/sample/a-buffer/main.ts
+++ b/sample/a-buffer/main.ts
@@ -85,6 +85,7 @@ const opaquePipeline = device.createRenderPipeline({
   layout: 'auto',
   vertex: {
     module: opaqueModule,
+    entryPoint: 'main_vs',
     buffers: [
       {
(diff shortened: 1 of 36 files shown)


beaufortfrancois:
@Djjanks You should update your browser to version 121. See https://developer.chrome.com/blog/new-in-webgpu-121#default_entry_points_to_shader_modules

ben-clayton:
These fields were removed with https://github.com/webgpu/webgpu-samples/pull/334, as the WebGPU specification was updated so that the entry point can be omitted if there is only one entry point in the WGSL with the requested stage.

Is your browser up to date?

Djjanks:
Thank you very much! I did have an issue during the update, but reinstalling resolved my problem. I apologize for the inconvenience caused by the merge request.

PR #423, opened 2024-06-15
Show a dialog if WebGPU isn't available or there's an error

kainino0x:
This won't catch JS exceptions in the middle of the actual sample, but it at least shows a nice message when WebGPU is not available for some reason, or there's a validation error (usually because the browser doesn't support the sample).

I would like to catch JS exceptions but it would be a bit more complicated. Thinking of just registering an onunhandledrejection handler and wrapping each sample in a `(async () => { ... })()` but that's a bit silly.

Anyway what do you think of this - is the DOM stuff too heavy handed for a sample util?

### DIFF
--- a/sample/a-buffer/main.ts
+++ b/sample/a-buffer/main.ts
@@ -1,6 +1,7 @@
 import { mat4, vec3 } from 'wgpu-matrix';
 import { GUI } from 'dat.gui';
 
+import { quitIfWebGPUNotAvailable } from '../util';
 import { mesh } from '../../meshes/teapot';
 
 import opaqueWGSL from './opaque.wgsl';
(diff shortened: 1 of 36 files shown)


greggman:
I'm not against this but my gut says this seems further from showing WebGPU?

What about something like

```
const adapter = await navigator.gpu?.requestAdapter();
const device = await adapter?.requestDevice();
if (!device) {
  showWebGPUIssues(adapter, device);  // from util.ts
}
```

Then the code shows more straight forward WebGPU in the example. You can still check if `navigator.gpu` is null, and/or if `adapter` was null, and/or if `device` was null or lost in `showWebGPUIssues`

That won't handle the case when asking for limits, features though.

kainino0x:
100% agree, I had some similar thoughts in the evening after I posted this. I think I can come up with something but it will have to wait a little bit.

kainino0x:
Finally revised this - PTAL

greggman (approved):
LGTM!

just random comments

* does it need a button? 

  The user clicks 'OK' for what purpose? I'm not saying to remove the button. Maybe the fact that it has one makes them more likely to read the message.

* not getting an adapter doesn't technically mean WebGPU is not available on this system. Compat for example, and in the future `fl2`. I can't think of a simple message though it's probably best as is.

kainino0x:
> * does it need a button?
>   The user clicks 'OK' for what purpose? I'm not saying to remove the button. Maybe the fact that it has one makes them more likely to read the message.

Not really, but sometimes there will be an error yet the sample keeps working - for example the reversedZ sample does this in Firefox right now. Also if some error comes up in an interactive demo like samplerParameters, it could be useful to dismiss it and change the settings back to the mode that didn't have an error.

> * not getting an adapter doesn't technically mean WebGPU is not available on this system. Compat for example, and in the future `fl2`. I can't think of a simple message though it's probably best as is.

Good point, I'll tweak it a little.

PR #428, opened 2024-06-18
Put search UI at top of editor

greggman:
Also, make max editor size to be 100vh

Hopefully that provides a better UX as it
keeps the search area on the screen.

It's not perfect because until the search UI hits the top of the screen it keeps moving so you have to chase it with your mouse. Putting at the bottom doesn't work because it's off the screen by default.

I spent a few minutes trying to make it hover at the bottom but failed.

Ultimately I'm mixed on the point. The code is on github. We could just have a link to the correct folder on github.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,6 +10,7 @@
       "license": "BSD-3-Clause",
       "dependencies": {
         "@codemirror/lang-javascript": "^6.2.2",
+        "@codemirror/search": "^6.5.6",
         "@codemirror/view": "^6.28.1",
         "@uiw/codemirror-theme-github": "^4.22.2",
         "@uiw/codemirror-theme-monokai": "^4.22.2",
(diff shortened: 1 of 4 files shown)


greggman:
before: Clicking in text area and pressing Ctrl-F/Cmd-F does nothing (because the search is at the bottom of the page). Even going to the bottom of the page, typing in "device" and picking "next" doesn't appear to do anyting.


https://github.com/webgpu/webgpu-samples/assets/234804/dcaaf8a1-c381-45de-9217-fdb72bbf3ba4

after: clicking in text area and pressing Ctrl-F/Cmd-F the search UI appears at the top. The text area is limited to 100vh so once it fill the screen you're scrolling through the editor, not the page. Typing "device" and pressing "next" takes you do each result.


https://github.com/webgpu/webgpu-samples/assets/234804/183152f8-5d88-43a0-93cb-f1a205d7bb4c

beaufortfrancois (approved):

PR #430, opened 2024-07-02
Add depth bias to wireframe sample

toji:
As was pointed out by @rconde01 in the matrix chat, the wireframe sample suffers from Z fighting that gives the wireframes a very aliased, sparkly appearance. This PR adds an adjustable `depthBias` and `depthBiasSlopeScale` slider that can be used both to improve the appearance of the wireframe and help visualize for users the effect those values have.

Before:
Screenshot 2024-07-02 103450

After:
Screenshot 2024-07-02 103528

Bias is applied to the lit mesh rather than the wireframe because apparently line and point primitives don't consider depth bias when rendering! Didn't know that previously. I've also added a checkbox to pause the scene's animation, as I found that helpful when specifically looking for the effects of the depth bias.

### DIFF
--- a/sample/wireframe/main.ts
+++ b/sample/wireframe/main.ts
@@ -5,6 +5,17 @@ import { randElement, randColor } from './utils';
 import solidColorLitWGSL from './solidColorLit.wgsl';
 import wireframeWGSL from './wireframe.wgsl';
 
+const settings = {
+  barycentricCoordinatesBased: false,
+  thickness: 2,
+  alphaThreshold: 0.5,
+  animate: true,
+  lines: true,
+  depthBias: 0.0,
+  depthBiasSlopeScale: 0.5,
+  models: true,
+};
+
 type TypedArrayView = Float32Array | Uint32Array;
 
 function createBufferWithData(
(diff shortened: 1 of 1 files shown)


greggman:
This is great! I actually tried adding that too and when I didn't see a difference I gave up for a while. I'm glad you figure out why it wasn't working for me.

Can you fix up the lint issues

toji:
Done!

sample/wireframe/main.ts
```diff
@@ -289,18 +323,15 @@ const renderPassDescriptor: GPURenderPassDescriptor = {
...
-  lines: true,
-  models: true,
-};
-
 const gui = new GUI();
 gui.add(settings, 'barycentricCoordinatesBased').onChange(addRemoveGUI);
 gui.add(settings, 'lines');
 gui.add(settings, 'models');
+gui.add(settings, 'animate');
+gui.add(settings, 'depthBias', -1, 1, 0.05).onChange(rebuildLitPipeline);
```
greggman, line 331: AFAICT, these are not needed to make the barycentric edge version work. The triangles are the same so the depth test of less-equal perfectly matches. Do these settings appearing in the barycentric case make that more confusing, as in, should these settings be removed and barycentric solution case?

greggman (approved):
lgtm with one question. up to you.

sample/wireframe/main.ts
toji: I wondered about that, but also observed that they DO have an effect in the barycentric mode (though as you pointed out aren't necessary to get a clean render) so I figured leaving them in was at the very least educational? I'm happy to suppress them in that case if you'd like.

greggman: I'm fine either way. Just don't want people to think they need to set them if they don't

PR #431, opened 2024-07-06
Add Occlusion Query Sample

greggman:
This is the simplest thing I could think of to
demonstrate how to use Occlusion Queries.


sample/occlusionQuery/main.ts
```diff
@@ -0,0 +1,342 @@
...
+   1,  1,  1,     0,  0,  1,
+  -1,  1,  1,     0,  0,  1,
+  -1, -1,  1,     0,  0,  1,
+   1, -1,  1,     0,  0,  1,
+  -1,  1, -1,     0,  0, -1,
+   1,  1, -1,     0,  0, -1,
+   1, -1, -1,     0,  0, -1,
+  -1, -1, -1,     0,  0, -1,
+]);
+const indices = new Uint16Array([
```
beaufortfrancois, line 186: I wonder if we could use something like `Array(24).fill(1).map((e,i)=>e+(i*1));`

sample/occlusionQuery/main.ts
```diff
@@ -0,0 +1,342 @@
...
+  depthStencil: {
+    depthWriteEnabled: true,
+    depthCompare: 'less',
+    format: depthFormat,
+  },
+});
+
+// prettier-ignore
+const cubePositions = [
+  { position: [-1,  0,  0], id: '🟥', color: 'red' },
```
beaufortfrancois, line 84: I believe css color names don't add much values to this sample. Using simply `rgba: [ 1, 0, 0, 1]` instead of `color: 'red'` would be enough and you could remove the `utils.ts` file

sample/occlusionQuery/main.ts
```diff
@@ -0,0 +1,342 @@
...
+});
+
+const occlusionQuerySet = device.createQuerySet({
+  type: 'occlusion',
+  count: objectInfos.length,
+});
+
+const resolveBuffer = device.createBuffer({
+  label: 'resolveBuffer',
+  size: objectInfos.length * 8,
```
beaufortfrancois, line 128: Maybe explain where does come from `* 8` here and below?
Or replace with `BigUint64Array.BYTES_PER_ELEMENT` to hint that we'll use BigUint64Array later

sample/occlusionQuery/main.ts
```diff
@@ -0,0 +1,342 @@
...
+  occlusionQuerySet,
+};
+
+const lerp = (a: number, b: number, t: number) => a + (b - a) * t;
+const lerpV = (a: number[], b: number[], t: number) =>
+  a.map((v, i) => lerp(v, b[i], t));
+const pingPongSine = (t: number) => Math.sin(t * Math.PI * 2) * 0.5 + 0.5;
+
+let depthTexture: GPUTexture | undefined;
+let results: BigUint64Array | undefined;
```
beaufortfrancois, line 228: I believe you don't need to declare `results` there right?

beaufortfrancois (commented):
I like this sample a lot! Than you for writing it!

sample/occlusionQuery/main.ts
greggman: that won't produce the correct indices. I think plain data is easier to follow of this. I reformatted the indices to be clearer.

sample/occlusionQuery/main.ts
greggman: changed.

sample/occlusionQuery/main.ts
greggman: `results` does need to be created there. `results` sticks around between frames because we don't know when `mapAsync` will resolve.

sample/occlusionQuery/main.ts
greggman: done

sample/occlusionQuery/main.ts
beaufortfrancois: We're redefining `results` when mapAsync resolves and only use its value there. So I assumed we could get rid of it and simply do the following: 
```js
    resultBuffer.mapAsync(GPUMapMode.READ).then(() => {
      const results = new BigUint64Array(resultBuffer.getMappedRange()).slice();
      resultBuffer.unmap();

      const visible = objectInfos
        .filter((_, i) => results[i])
        .map(({ id }) => id)
        .join('');
      info.textContent = `visible: ${visible}`;
    });
```

sample/occlusionQuery/main.ts
beaufortfrancois: Thank you!

sample/occlusionQuery/main.ts
greggman: you're right. fixed

beaufortfrancois (approved):
LGTM

Thanks addressing nits!

PR #432, opened 2024-07-10
Add HDR settings to Particles sample

beaufortfrancois:
Results:
- Works on Chrome Canary with `--enable-blink-features=WebGPUHDR`
- Flickers a lot on Safari Tech Preview with experimental WebGPU support
- Did not work on Firefox Nightly before this PR


beaufortfrancois:
@ccameron-chromium Please have a look as well

sample/particles/main.ts
```diff
@@ -330,6 +336,11 @@ const simulationUBOBuffer = device.createBuffer({
 const gui = new GUI();
 gui.add(simulationParams, 'simulate');
 gui.add(simulationParams, 'deltaTime');
+const hdrFolder = gui.addFolder('HDR Settings');
+hdrFolder
+  .add(simulationParams, 'toneMappingMode', ['standard', 'extended'])
+  .onChange(configureContext);
+hdrFolder.add(simulationParams, 'brightnessFactor', 0, 4, 0.1);
```
greggman, line 344:
```suggestion
hdrFolder.open();
```

wdyt? this will start with folder open

beaufortfrancois:
> I think adding a HDR sample is great but I wonder if the particle example is a good example for it vs some dedicated sample? How will people know to go look at "particles" (guess you could rename it particles(HDR)?)

I've added the HDR bit to make it clearer. Thanks for the suggestion!

> 
> Also, I couldn't see any difference at a glance. I ran with
> 
> ```
> /Applications/Google\ Chrome\ Canary.app/Contents/MacOS/Google\ Chrome\ Canary --use-data-dir=$T/delme-canray-6933 --enable-unsafe-webgpu --enable-webgpu-developer-features --enable-blink-features=WebGPUHDR
> ```

I tried your command line and it works great on the Built-in Liquid Retina XDR Display of my Apple M1 Pro device. When plugged to an external non-HDR display, there's no difference as the display doesn't support HDR at all as expected. Note that you only need `/Applications/Google\ Chrome\ Canary.app/Contents/MacOS/Google\ Chrome\ Canary --enable-blink-features=WebGPUHDR`.

> So it makes me wonder if rather than a toggle, whichever HDR sample we end up with (could be this one) shows 2 canvases. One HDR, one not, side by side. Then, that they are different would stick out?

I think the particles sample really shines with HDR. Traditional HDR demos like https://webkit.org/blog-files/color-gamut/ are "okay" to understand what is HDR but I think the webgpu-samples should focus on how the WebGPU API works in regards to HDR with `toneMapping` and `format`.

sample/particles/index.html
```diff
@@ -3,7 +3,7 @@
   <head>
     <meta charset="utf-8">
     <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
-    <title>webgpu-samples: particles</title>
+    <title>webgpu-samples: particles (HDR)</title>
```
kainino0x, line 6: We would need the "HDR" to appear in the sidebar too, so I think we'd want to rename the whole folder to `sample/particlesHDR`

sample/particles/main.ts
```diff
@@ -24,13 +24,16 @@ const context = canvas.getContext('webgpu') as GPUCanvasContext;
...
 canvas.width = canvas.clientWidth * devicePixelRatio;
 canvas.height = canvas.clientHeight * devicePixelRatio;
-const presentationFormat = navigator.gpu.getPreferredCanvasFormat();
-
-context.configure({
-  device,
-  format: presentationFormat,
-  alphaMode: 'premultiplied',
-});
+const presentationFormat = 'rgba16float';
```
kainino0x, line 27: What is causing flickering? Is it this? Or has safari implemented .toneMapping.mode already?

sample/particles/particle.wgsl
```diff
@@ -125,6 +126,9 @@ fn simulate(@builtin(global_invocation_id) global_invocation_id : vec3u) {
     let uv = vec2f(coord) / vec2f(textureDimensions(texture));
     particle.position = vec3f((uv - 0.5) * 3.0 * vec2f(1.0, -1.0), 0.0);
     particle.color = textureLoad(texture, coord, 0);
+    particle.color.r *= sim_params.brightnessFactor;
+    particle.color.g *= sim_params.brightnessFactor;
+    particle.color.b *= sim_params.brightnessFactor;
```
kainino0x, line 131: FWIW - I haven't tested this but - I think it may look better if the colors are HDR and the white "WebGPU" text is not (it makes sense for text to be "paper-white").

sample/particles/main.ts
beaufortfrancois: `rgba16float` is what's causing the issue from what I can tell.

kainino0x: Makes sense, thanks. I think it's OK to land the sample change while Safari still has the bug, but FYI @mwyrzykowski

mwyrzykowski: Yes please go ahead and land it 👍 we haven't extensively tested float16 canvas support

mwyrzykowski (approved):

sample/particles/main.ts
mwyrzykowski: @beaufortfrancois out of curiosity which version of Safari Tech Preview do you see the flickering? I pulled your changes and ran particles locally, it doesn't flicker for me, but I am trying ToT WebKit. Was it Safari Technology Preview 198?

beaufortfrancois: I was able to make it flicker on STP Release 194 (Safari 17.4, WebKit 19619.1.11.111.2). It does flicker anymore on STP Release 199 (Safari 18.0, WebKit 19619.1.22.5). We're good then!

sample/particles/particle.wgsl
beaufortfrancois: I'd recommend you give it a try locally. I personally like it the way it is but I welcome changes ;)

sample/particles/index.html
beaufortfrancois: Instead of renaming the whole folder (and losing link URL), I simply updated `samples.ts`. WDYT?

sample/particles/main.ts
mwyrzykowski: Oh yes there was a known flickering bug, I think it was fixed around STP 197 or so, thanks for confirming!

sample/particles/particle.wgsl
kainino0x: No worries, I trust your judgement!

sample/particles/index.html
kainino0x: didn't realize we could do that, LGTM

kainino0x (approved):

src/samples.ts
```diff
@@ -116,7 +116,7 @@ export const pageCategories: PageCategory[] = [
       normalMap,
       shadowMapping,
       deferredRendering,
-      particles,
+      'particles (HDR)': particles,
```
kainino0x, line 119: Unfortunately this didn't quite work, https://webgpu.github.io/webgpu-samples/?sample=particles doesn't link correctly anymore. I'll try a fix

kainino0x: fix in #437

mwyrzykowski:
Just tried it out on a MBP - works great! 🎉

beaufortfrancois:
I've tested Android as well and it works great there also! 
@mwyrzykowski Does https://webgpu.github.io/webgpu-samples/?sample=particles also work on iOS?

mwyrzykowski:
It does but WebKit doesn’t support HDR canvas yet. But it appears even on an iPhone 15 pro the sample thinks HDR is not supported:

IMG_1336

beaufortfrancois:
I'm using `window.matchMedia('(dynamic-range: high)')` to test whether the display is HDR compatible or not. https://developer.mozilla.org/en-US/docs/Web/CSS/@media/dynamic-range#browser_compatibility says it should be supported on Safari on iOS 🤷 .

mwyrzykowski:
> I'm using `window.matchMedia('(dynamic-range: high)')` to test whether the display is HDR compatible or not. https://developer.mozilla.org/en-US/docs/Web/CSS/@media/dynamic-range#browser_compatibility says it should be supported on Safari on iOS 🤷 .

Oh interesting, I reloaded the page and the 'display not compatible' banner disappeared, looks like the detection works

beaufortfrancois:
> It does but WebKit doesn’t support HDR canvas yet. 

Is there a way to detect in JavaScript whether HDR canvas is supported instead of using this media query?

mwyrzykowski:
> Is there a way to detect in JavaScript whether HDR canvas is supported instead of using this media query?

Apologies for the delay but I looked around and didn't see anything conclusive. Could we check for `context.colorMetadata  === undefined` as that appears required by the HDR canvas specification? https://github.com/w3c/ColorWeb-CG/blob/main/hdr_html_canvas_element.md#add-hdr-rendering-behavior-and-hdr-metadata-to-canvasrenderingcontext2dsettings

Or should one be able to use HDR in WebGPU without an higher bit depth canvas (e.g., R10G10B10A2 type formats)?

beaufortfrancois:
`context.colorMetadata` is not implemented in Chromium sadly.

And `GPUCanvasConfiguration` is a dictionary, not an interface, so I can't check without calling `configure()` first whether `toneMapping` is implemented.

I wonder if I should leave it as is.

mwyrzykowski:
It’s certainly fine with me to leave as-is 👍

PR #435, opened 2024-07-15
Enclose all URLs in descriptions with <> syntax

kainino0x:
This markdown syntax guarantees the link will be correctly detected, instead of relying on autodetection (which could miss some URLs, pick up adjacent punctuation, etc.)


toji (approved):
Huh, didn't know about this Markdown syntax! LGTM.

PR #443, opened 2024-09-10
Some mobile browsers require playsInline to be set to true for the video

mwyrzykowski:
Some mobile browsers require playsInline to be set to true for the video to play with the rest of the content.

To avoid WebGPU samples testing differences between mobile and desktop browsers, set playsInline = true prior to playing the video.

### DIFF
--- a/sample/videoUploading/main.ts
+++ b/sample/videoUploading/main.ts
@@ -10,6 +10,7 @@ quitIfWebGPUNotAvailable(adapter, device);
 // Set video element
 const video = document.createElement('video');
 video.loop = true;
+video.playsInline = true;
 video.autoplay = true;
 video.muted = true;
 video.src = '../../assets/video/pano.webm';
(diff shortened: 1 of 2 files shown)


mwyrzykowski:
cc @kainino0x @greggman @beaufortfrancois who recently modified this file, are you ok with this change?

mwyrzykowski:
> LGTM overall. I found two results in https://github.com/search?q=repo%3Awebgpu%2Fwebgpu-samples%20createElement(%27video%27)&type=code but I think [`sample/videoUploading/video.ts`](https://github.com/webgpu/webgpu-samples/blob/a7c8b6867c495ff145f34772b105ede882340a68/sample/videoUploading/video.ts#L8) is not used anymore

Sure let's update the other occurrence too even though I agree, it seems unused.

kainino0x (approved):
LGTM. Doesn't look like we are using `playsInline` on CTS, we might want it there too?

PR #445, opened 2024-09-20
alphaToCoverage: show resolve result; draw twice with different colors/alphas

kainino0x:
This demonstrates how different alpha-to-coverage algorithms produce different blending results.

image


toji (approved):
LGTM! Seems like a good improvement on the initial version of the sample.

PR #452, opened 2024-10-01
Use transparency in helloTriangle*, remove alphaMode on rest

kainino0x:
None of the samples used transparent canvases, so they shouldn't set alphaMode.
We should demonstrate transparent canvases though, so I made the two helloTriangle samples do that.

Technically opaque canvases force an alpha clear on Mac, but probably we should stick to the spec default. Someday maybe we can do something about that alpha clear.

### DIFF
--- a/sample/alphaToCoverage/main.ts
+++ b/sample/alphaToCoverage/main.ts
@@ -69,7 +69,6 @@ const context = canvas.getContext('webgpu') as GPUCanvasContext;
 context.configure({
   device,
   format: presentationFormat,
-  alphaMode: 'premultiplied',
 });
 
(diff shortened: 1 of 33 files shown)


toji (approved):
I think it's a good idea to take the `alphaMode` off of the majority of the samples, but I feel like we could use a slightly better demonstration of transparent canvases than the hello triangles? As-is it would be difficult to tell that they are transparent vs. simply clearing to a color that matches the background.

I'd prefer to have a sample that shows some DOM content (even just simple text) behind the canvas. Happy to help build it!

In the meantime, I think this can land as-is.

PR #455, opened 2024-10-02
Fix toneMapping reference

greggman:
AFAICT, the spec requires toneMapping period. Even if the browser doesn't support toneMapping.mode = 'extended' it requires it still requires it to exist. It will just be set to `'standard'` if the implementation does not support `'extended'`.

The only other case is when the canvas has not been configured in which case `getConfiguration` returns undefined.

Both `getConfiguration` and `toneMapping` were added to the spec close enough that any implemetation that supports `getConfiguration` should be requried to include `toneMapping`.

### DIFF
--- a/sample/particles/main.ts
+++ b/sample/particles/main.ts
@@ -355,7 +355,7 @@ function getHdrFolderName() {
   }
   if (
     simulationParams.toneMappingMode === 'extended' &&
-    context.getConfiguration().toneMapping?.mode !== 'extended'
+    context.getConfiguration().toneMapping.mode !== 'extended'
   ) {
     return "HDR settings ⚠️ Browser doesn't support HDR canvas";
   }


beaufortfrancois:
You're right. See https://github.com/webgpu/webgpu-samples/pull/451#discussion_r1783344030

beaufortfrancois:
@kainino0x I believe we can merge this PR. What do you think?

beaufortfrancois (approved):

kainino0x:
Sorry I forgot about this PR because I replied to the spec issue and PR instead.
According to the (open) spec PR the `?` is still needed because `toneMapping` is the thing that won't exist in a browser that doesn't support HDR WebGPU (assuming Safari and Firefox pass through that point before supporting it).
https://github.com/gpuweb/gpuweb/pull/4922/files

greggman:
> According to the (open) spec PR the `?` is still needed because `toneMapping` is the thing that won't exist in a browser that doesn't support HDR WebGPU (assuming Safari and Firefox pass through that point before supporting it).

Can we just require `toneMapping`? Maybe it's minor but one path is "forgot or didn't know I needed the `?` and my app crashes with `TypeError: Cannot read properties of undefined`" and the other just returns `undefined` so code checking `getConfigure().toneMapping.mode` at least doesn't crash.

`getConfigure()` hasn't shipped to stable so we still have time

kainino0x:
Require `toneMapping` and have `toneMapping.mode` be the thing that's missing on browsers that don't implement it? Would be fine with me.

beaufortfrancois:
> `getConfigure()` hasn't shipped to stable so we still have time

FYI Chrome always returns `toneMapping` in `getConfiguration()` if configuration is not null. See https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/modules/webgpu/gpu_canvas_context.cc;l=622;drc=a7b08979892095de2548e193c95e93a01ed8b214

PR #457, opened 2024-10-07
Add a Marching Cubes external sample.

tcoppex:
Hi,

This sample demonstrates how to dynamically generate procedural meshes based on a signed distant function using a multi-pass marching cubes algorithm on the GPU. It also demonstrates tri-planar mapping and integrates an Ambient Occlusion calculation in the final vertex generation stage.

Live preview: https://tcoppex.github.io/webgpu-marchingcubes/

Cheers.

### DIFF
--- a/sample/marchingCubes/meta.ts
+++ b/sample/marchingCubes/meta.ts
@@ -1,0 +1,10 @@
+export default {
+  name: 'Marching Cubes',
+  description: `This example demonstrates how to dynamically generate procedural meshes using a signed distance field and a multi-pass marching cubes algorithm on the GPU.
+
+Source at <https://github.com/tcoppex/webgpu-marchingcubes>
+`,
+  filename: __DIRNAME__,
+  url: 'https://tcoppex.github.io/webgpu-marchingcubes/',
+  sources: [],
+};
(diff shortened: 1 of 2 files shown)


kainino0x:
This is a nice sample! Unfortunately it is having issues on Mac :(
Most likely some kind of undefined behavior (out of bounds?) but let us know if it looks like a Chrome bug.
Screenshot 2024-10-09 at 11 52 23 AM

tcoppex:
Hi @kainino0x,

Thanks for your feedback! It does seem like an out-of-bounds issue. So far I've only tested it on Chrome v129.x on Windows 11, as browser builds don't appear to be compliant yet on GNU/Linux. I'll get back to you if I find anything. In the meantime, I'll mark the PR as a draft.

Thanks.

tcoppex:
While I'm not 100% sure why it worked, the issue should now be resolved on Mac. 

I reduced the size of the density volume, which mainly affected the compute workgroup size. Despite it was still within general device limits (as defined by https://web3dsurvey.com/), it has done the trick.

kainino0x (approved):
Nice! Thanks for the sample!

PR #459, opened 2024-10-14
Remove gif.js from index.html

beaufortfrancois:
image


kainino0x (approved):
oops, my bad! Thought I left that in only on the branch.

PR #460, opened 2024-10-14
Fix fragment shader path

beaufortfrancois:
image


toji (approved):
Ah, thanks for fixing that!

PR #464, opened 2024-10-16
Update chokidar

greggman:
Note: This is dependent on https://github.com/webgpu/webgpu-samples/pull/463

Update chokidar to v4

chokidar v3 is no longer supported. v4 requires node 20 so

1. Added a check for v20+

2. Make script exit if any spawn gets an error.

   This required watching for processes to exit and
   killing the other processes.

3. Update to use glob as chokidar v4 no longer does it

Also fixed a minor warning in sample/utis.ts


sample/util.ts
```diff
@@ -22,6 +22,7 @@ export function quitIfWebGPUNotAvailable(
   if (!device) {
     quitIfAdapterNotAvailable(adapter);
     fail('Unable to get a device for an unknown reason');
+    return;
```
kainino0x, line 25: IIRC `fail` should always throw, so this should be dead code?

build/tools/serve.js
```diff
@@ -3,17 +3,29 @@ import { mkdirSync } from 'fs';
 
 mkdirSync('out', { recursive: true });
 
-spawn('npm', ['run', 'watch'], {
+const spawns = [];
+
+function spawnAndCheck(cmd, args, options) {
+  const s = spawn(cmd, args, options);
+  spawns.push(s);
```
kainino0x, line 10: doesn't seem like items are ever removed from this array. If that's not a problem, add a comment about why.

sample/util.ts
greggman: Without this I get these errors from the lines below.

```
(!) [plugin typescript] sample/util.ts (27:3): @rollup/plugin-typescript TS18047: 'device' is possibly 'null'.
/Users/gman/src/webgpu-samples/sample/util.ts:27:3

27   device.lost.then((reason) => {
     ~~~~~~
```

build/tools/serve.js
greggman: done

sample/util.ts
kainino0x: ahh, that makes sense, this is probably too complex of a scenario to trigger type narrowing on `device` even though `fail` presumably has a return type of `never`.

build/tools/serve.js
```diff
@@ -9,9 +9,12 @@ function spawnAndCheck(cmd, args, options) {
   const s = spawn(cmd, args, options);
   spawns.push(s);
   s.on('close', (code) => {
-    console.log(cmd, 'exited with code:', code);
-    spawns.forEach((s) => s.kill());
-    process.exit(code);
+    spawns.splice(spawns.indexOf(s), 1);
```
kainino0x, line 12: nit: maybe `spawns` could be a `Set`? but not important

kainino0x (approved):

Issue #458, opened 2024-10-08
Address Particles HDR sample blue trail


beaufortfrancois:
The following blue trail is a pet peeve of me in the [Particles (HDR) sample](https://webgpu.github.io/webgpu-samples/?sample=particles):

![Screenshot 2024-10-08 at 10 17 46](https://github.com/user-attachments/assets/9144d4d4-8f97-4d53-8485-4e4291d54a24)

I can reproduce on macOS (Apple M1 Pro device with Chrome and Safari), but not on Android.
Could this be a Metal driver issue or something else? FYI @mwyrzykowski

mwyrzykowski:
I have also noticed this and it seems to occur in different locations in Chrome (left, trail in far top-left) compared to Safari (right, trail in middle top-left)
<img width="836" alt="Screenshot 2024-10-08 at 8 11 06 AM" src="https://github.com/user-attachments/assets/35e92416-cde3-4f13-8778-a0cfbb293be6">

I have not debugged the sample first, so I would start there before thinking it is a metal bug. The issue is present on AMD Intel Mac as well, so if it is a Metal issue it is not hardware specific.

mwyrzykowski:
that looks better:
<img width="1238" alt="Screenshot 2024-10-08 at 8 23 50 AM" src="https://github.com/user-attachments/assets/12a3b107-5a3a-4aa2-b600-0e6a10b47ab2">

It is an issue in the sample, I don't reproduce when I disable fastMath (`options.fastMathEnabled = NO;`) in WebKit

The WGSL specification is supposed to allow for variances due to having fast math enabled, so likely these functions:
```
fn init_rand(invocation_id : u32, seed : vec4f) {
  rand_seed = seed.xz;
  rand_seed = fract(rand_seed * cos(35.456+f32(invocation_id) * seed.yw));
  rand_seed = fract(rand_seed * cos(41.235+f32(invocation_id) * seed.xw));
}

fn rand() -> f32 {
  rand_seed.x = fract(cos(dot(rand_seed, vec2f(23.14077926, 232.61690225))) * 136.8168);
  rand_seed.y = fract(cos(dot(rand_seed, vec2f(54.47856553, 345.84153136))) * 534.7645);
  return rand_seed.y;
}
```
are making assumptions about numerical accuracy which are not guaranteed by WGSL.

greggman:
[...]

PR #469, opened 2024-10-22
Fix blue trail

beaufortfrancois:
FIX https://github.com/webgpu/webgpu-samples/issues/458

### DIFF
--- a/sample/particles/particle.wgsl
+++ b/sample/particles/particle.wgsl
@@ -5,8 +5,8 @@ var<private> rand_seed : vec2f;
 
 fn init_rand(invocation_id : u32, seed : vec4f) {
   rand_seed = seed.xz;
-  rand_seed = fract(rand_seed * cos(35.456+f32(invocation_id) * seed.yw));
-  rand_seed = fract(rand_seed * cos(41.235+f32(invocation_id) * seed.xw));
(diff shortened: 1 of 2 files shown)


sample/particles/particle.wgsl
```diff
@@ -5,8 +5,8 @@ var<private> rand_seed : vec2f;
 
 fn init_rand(invocation_id : u32, seed : vec4f) {
   rand_seed = seed.xz;
-  rand_seed = fract(rand_seed * cos(35.456+f32(invocation_id) * seed.yw));
-  rand_seed = fract(rand_seed * cos(41.235+f32(invocation_id) * seed.xw));
+  rand_seed = fract(rand_seed * cos(35.456+f32(invocation_id) * seed.yw * 0.01f));
+  rand_seed = fract(rand_seed * cos(41.235+f32(invocation_id) * seed.xw * 0.01f));
```
kainino0x, line 9: Seems like this is probably only necessary because seed.x and seed.y can be large:
https://github.com/webgpu/webgpu-samples/blob/1e3c2d387d52918295b91006256fc36e48d095b4/sample/particles/main.ts#L416-L419

Could we just adjust the seed instead so it's not so large? I'm not sure what effect it would have.

kainino0x:
Oh! Fantastic!

kainino0x (approved):

Issue #410, opened 2024-03-28
Indirect Draw Call Sample


cmhhelgeson:
Should there be an example that demonstrates best practices for using indirect draw calls, ala the renderBundles example? Or perhaps indirect draw calls could be an option in a separate sample?

kainino0x:
@toji did you have a demo doing some interesting things with indirect draws?

greggman:
AFAIK this sample https://webgpu.github.io/webgpu-samples/?sample=bundleCulling uses indirect draw calls.

https://github.com/toji/webgpu-bundle-culling/blob/3098596aef18acd91e93d85156b49f08fcee9831/index.html#L694

toji:
Yeah, the bundle culling code does use indirect draws, and I talk about their usage in that sample here: https://toji.dev/webgpu-best-practices/render-bundles#indirect-draws (Plus some additional optimization information here: https://toji.dev/webgpu-best-practices/indirect-draws) For what it's worth https://webgpu.github.io/webgpu-samples/?sample=metaballs also uses indirect draws, and is probably a more straightforward use case for them. They're both reasonably complex samples, though, and focused on techniques other than the indirect part.

It may be worthwhile to create a simpler sample focused specifically on indirect draws. It could be as basic as a variant of the spinning cube that uses an indirect draw instead of a traditional one, but that doesn't really touch on the "why" of indirect drawing. A sample with some simple generative geometry (like the metaballs, but ideally even more basic) would serve as a better illustration but it's easy for that to get overly complex fast.

PR #471, opened 2024-10-30
Use drawIndirect instead of draw in particles sample

beaufortfrancois:
As discussed in https://github.com/webgpu/webgpu-samples/issues/410, a sample featuring drawIndirect would be useful to developers. This PR attempts to do that by replacing `draw()` with `drawIndirect()` in the particles sample. Let me know what you think.

### DIFF
--- a/sample/particles/main.ts
+++ b/sample/particles/main.ts
@@ -398,6 +398,11 @@ const computeBindGroup = device.createBindGroup({
   ],
 });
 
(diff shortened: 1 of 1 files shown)


kainino0x:
IMO this makes it look like this is a best practice, but it isn't. We should either have a sample that uses it meaningfully (for some kind of culling?) or just a simple triangle drawn with drawIndirect just to show how you call it.

PR #472, opened 2024-11-12
Add timestamp query example

eliemichel:
Add an example that issues a single draw call of a cube (taken from the `texturedCube` example) and measures its duration thanks to timestamp queries.


sample/timestampQuery/main.ts
```diff
@@ -0,0 +1,307 @@
...
+  if (!hasOngoingTimestampReadback) {
+    hasOngoingTimestampReadback = true;
+    timestampMapBuffer
+      .mapAsync(GPUMapMode.READ, 0, timestampBufferSize)
+      .then(() => {
+        const buffer = timestampMapBuffer.getMappedRange(0, timestampBufferSize);
+        const timestamps = new BigUint64Array(buffer);
+
+        // Measure difference (in bigints)
+        const elapsedNs = timestamps[1] - timestamps[0];
```
kainino0x, line 292: Per spec here: https://gpuweb.github.io/gpuweb/#timestamp
we should drop any samples where `elapsedNs` is negative.
(And have a comment linking to the spec to explain why)

sample/timestampQuery/main.ts
```diff
@@ -0,0 +1,307 @@
...
+    },
+  ],
+});
+
+const renderPassDescriptor: GPURenderPassDescriptor = {
+  colorAttachments: [
+    {
+      view: undefined, // Assigned later
+
+      clearValue: [0.5, 0.5, 0.5, 1.0],
```
kainino0x, line 165: Red on gray is kind of hard on the eyes :)
Maybe red on black?

sample/timestampQuery/main.ts
```diff
@@ -0,0 +1,307 @@
+import { GUI } from 'dat.gui';
```
kainino0x, line 1: unused

src/samples.ts
```diff
@@ -171,6 +172,7 @@ export const pageCategories: PageCategory[] = [
     samples: {
       animometer,
       workloadSimulator,
+      timestampQuery,
```
kainino0x, line 175: I think I would categorize this under "WebGPU Features" since it's useful as sample code, more so than as just a benchmark.

sample/timestampQuery/main.ts
kainino0x, line 1: Right now it's actually a bit hard to sift out the parts of this code that are relevant reference code for timestamp queries. For this I have two possible ideas:
- Split as much as possible of the timestamp-related stuff into a separate file so that readers can easily see the two halves separately.
- Add this onto an existing sample (still as a separate file). If we do this we'd need to highlight it in the sidebar, similar to how we put "(HDR)" on the particles demo. This would be a bit harder to find though, and probably not categorized in the right section of the sidebar.

sample/timestampQuery/main.ts
```diff
@@ -0,0 +1,307 @@
...
+} from '../../meshes/cube';
+
+import basicVertWGSL from '../../shaders/basic.vert.wgsl';
+import sampleTextureMixColorWGSL from '../../shaders/red.frag.wgsl';
+import { quitIfWebGPUNotAvailable, fail } from '../util';
+
+const canvas = document.querySelector('canvas') as HTMLCanvasElement;
+const adapter = await navigator.gpu?.requestAdapter();
+if (!adapter.features.has("timestamp-query")) {
+  fail('WebGPU timestamp queries are not supported on this system');
```
kainino0x, line 19: With some tweaking this could also make a good example of using a feature when it's available, and falling back when it's not. Not sure if we have any right now.

I think it would actually be more useful as a reference for timestamp queries too, because almost any shipping code that uses timestamp queries should also work when they're unavailable.

sample/timestampQuery/main.ts
eliemichel: It was indeed confused, but not sure how engineered those examples should be. I split the code and moved most of the timestamp logic into a dedicated `TimestampQueryManager.ts` that one should be able to easily reuse as a base.

sample/timestampQuery/main.ts
eliemichel: The TimestampQueryManager should handle this graceful degradation now!

eliemichel:
Thanks for the feedback! 🙏

sample/timestampQuery/TimestampQueryManager.ts
```diff
@@ -0,0 +1,98 @@
...
+        const rawData = buffer.getMappedRange(0, buffer.size);
+        const timestamps = new BigUint64Array(rawData);
+
+        onTimestampReadBack(timestamps);
+
+        buffer.unmap();
+        this.hasOngoingTimestampReadback = false;
+      })
+    });
+  }
```
kainino0x, line 97: `resolve` is never called so this promise will never resolve. This function is kind of mixing callback and promise styles, but the returned promise is not used anyway so it could be removed.

```suggestion
  readAsync(onTimestampReadBack: (timestamps: BigUint64Array) => void): void {
    if (!this.timestampSupported) return;
    if (this.hasOngoingTimestampReadback) return;

    this.hasOngoingTimestampReadback = true;

    const buffer = this.timestampMapBuffer;
    void buffer.mapAsync(GPUMapMode.READ)
      .then(() => {
        const rawData = buffer.getMappedRange();
        const timestamps = new BigUint64Array(rawData);

        onTimestampReadBack(timestamps);

        buffer.unmap();
        this.hasOngoingTimestampReadback = false;
      });
  }
```

also nit: the `0, buffer.size` aren't needed because those are the defaults in both mapAsync and getMappedRange

kainino0x (commented):
Split LGTM, just one comment!

sample/timestampQuery/TimestampQueryManager.ts
eliemichel: Ow my bad, the whole Promise thing is a leftover from a previous attempt where readAsync was returning a Promise **instead** of taking a callback argument, but it was not possible because of mapping lifetime!

kainino0x (approved):

kainino0x:
Thank you!

beaufortfrancois:
As you can see in https://developer.chrome.com/blog/new-in-webgpu-121#timestamp_queries_in_compute_and_render_passes and https://webgpu.github.io/webgpu-samples/?sample=computeBoids, there was already a sample featuring timestamp-query.

Note that since we moved to a white background iframe, the text on large desktop didn't show up properly. I'm fixing it with https://github.com/webgpu/webgpu-samples/pull/477

greggman:
> Note that since we moved to a white background iframe,

We did not move to a white background iframe. It's was a bug in Chrome. It's not fixed still but there's a workaround https://github.com/webgpu/webgpu-samples/pull/478

eliemichel:
> As you can see in https://developer.chrome.com/blog/new-in-webgpu-121#timestamp_queries_in_compute_and_render_passes and https://webgpu.github.io/webgpu-samples/?sample=computeBoids, there was already a sample featuring timestamp-query.

The journey I followed is (a) "oh I need to quickly test a thing with timestamp queries" (b) "mmh I remember there is this nice sample repo where I can find examples, let's look for one about timestamps" (c) "arf, don't find an example whose name says "timestamp"... maybe in benchmark section? nope" (d) "okey then, let's write it then and share it back where I expected to find it". :)

beaufortfrancois:
I apologize if my previous message came across as harsh. That was not my intention. I realize that I could have been clearer in the title by specifying that it involved timestamp queries. I take full responsibility for this oversight.

beaufortfrancois:
> > As you can see in https://developer.chrome.com/blog/new-in-webgpu-121#timestamp_queries_in_compute_and_render_passes and https://webgpu.github.io/webgpu-samples/?sample=computeBoids, there was already a sample featuring timestamp-query.
> 
> The journey I followed is (a) "oh I need to quickly test a thing with timestamp queries" (b) "mmh I remember there is this nice sample repo where I can find examples, let's look for one about timestamps" (c) "arf, don't find an example whose name says "timestamp"... maybe in benchmark section? nope" (d) "okey then, let's write it then and share it back where I expected to find it". :)

I apologize if my previous message came across as harsh. That was not my intention. I realize that I could have been clearer in the title by specifying that it involved timestamp queries. I take full responsibility for this oversight.

eliemichel:
> I apologize if my previous message came across as harsh. That was not my intention. I realize that I could have been clearer in the title by specifying that it involved timestamp queries. I take full responsibility for this oversight.

No worry, really :)

PR #473, opened 2024-11-14
Add "view as standalone webpage" link; improvements for external samples

kainino0x:
- "View as a standalone webpage" does the same thing as "open in new tab" on the sidebar links.
- Fixed the sidebar links for external samples
- Made external sample source part of the metadata instead of putting it in each sample's description
- Highlight broken links (was mostly just useful for me while doing this change)

image

image


greggman (approved):
lgtm

PR #475, opened 2024-11-14
Avoid superluminant value in transparentCanvas

kainino0x:
Ugh, this is really easy to do accidentally. Got lucky and noticed that it rendered incorrectly on one of my displays (but not the other!)

https://gpuweb.github.io/gpuweb/#out-of-gamut-premultiplied-rgba-value


greggman (approved):

toji:
Thanks, that was a leftover from debugging. 🙄

PR #476, opened 2024-11-15
Cleanup timestamp-query example

greggman:
This was a great example. I started this change as just removing a `console.log` that seemed like it shouldn't be there but then I noticed the code didn't actually handle the case when 'timestamp-query' doesn't exist and ended up making a bunch of changes.

I didn't mean to step on anyone's toes. Here's the change though.

1. Remove `console.log` from timestamp query sample

   This was just filling all of memory with strings. Probably not a big deal since no one is likely to run the sample for hours but still....

   Maybe add a `<pre>` spot that holds the last N queries if you want to see a history?

2. no need for `hasOngoingTimestampReadback`

   `GPUBuffer` already has `mapState` which tells you this state so no need to track it separately.

3. There's no need for timestampCount

   It's already on `GPUQuerySet.count`

4. code didn't check if timestamps are supported before adding `timestamp` section to the render pass so the sample got a validation error if timestamp-query was not avaialble.

   Note: you can test this with the webgpu-dev-extension. Type 'timestamp-query' into the "Block Feature" field and reload the page.

5. Refactor not to use `readAsync`

   This seems like a problematic API. You call `readAsync` and if the map is pending then your callback is ignored. Ideally, if a callback was the right API design, then every callback you pass should get called and if it's pending then it would need to add your callback to a list of callbacks to be called when it's ready.

   Further, the timestamps passed back are only valid during the callback. If you tried to keep them they'd magically disappear as soon as your callback exited since they're directly the mapped buffer which will be unmapped as soon as you exit.

   Further, I'd argue for this sample at least, all you want is the last timing.

   So, changed the API to just give you the last times. No callbacks.

6. It's not clear what TimestampQueryManager was really managing.

   It wasn't adding the timestamps for you. You had to add them yourself and pull out the timestamp query object from internal fields inside TimestampQueryManager.

   Further, because you were adding them yourself you'd have to check yourself, if queries were supported then add them.

   Then, beacuse you were adding them yourself it was up to you to choose start and ends indices.

   It was also passing out BigInt but BitInt is hard to use correctly. I'd argue you don't need it. IIUC, Number.MAX_SAFE_INTEGER is 104 days of nanoseconds so it's fine to convert and give the user something easy to work with.

   So, I changed it to take a pairId and then use that to set both start and end. That way it can auto-subtract the pairs and give you Number. The user doesn't have to deal with BigInt. It still has the issue the timestamps a only kind of valid during the callback but that's not really the contract. Rather, they are always the last read value but at least they don't go to undefined if you happened to save timestamp?

Other: I considered even getting rid of `pairId` and have `addTimestampWrite` just auto increment the indices to use. It would then reset to 0 when you call `update`. With that, it could easily add up the times and give you the total time across passes. That way you wouldn't have to manage the pairIds yourself. On the otherhand, if you wanted individual pass timings you need to know which pass got which pairId. It could pass the pairId back from `addTimestamp` but I was less sure about that change.

Anyway, I hope this is considered an improvement.

One other thing, I don't understand `PerfCounter`. If you get one bad outsized timing it seems like it will take an extremely long time to recover though maybe I don't understand the math.

### DIFF
--- a/sample/timestampQuery/TimestampQueryManager.ts
+++ b/sample/timestampQuery/TimestampQueryManager.ts
@@ -4,9 +4,6 @@ export default class TimestampQueryManager {
   // class does nothing.
   timestampSupported: boolean;
 
-  // Number of timestamp counters
-  timestampCount: number;
-
   // The query objects. This is meant to be used in a ComputePassDescriptor's
   // or RenderPassDescriptor's 'timestampWrites' field.
   timestampQuerySet: GPUQuerySet;
@@ -17,37 +14,51 @@ export default class TimestampQueryManager {
   // A buffer to map this result back to CPU
   timestampMapBuffer: GPUBuffer;
 
-  // State used to avoid firing concurrent readback of timestamp values
-  hasOngoingTimestampReadback: boolean;
+  // Last times
+  timestamps: number[];
 
   // Device must have the "timestamp-query" feature
-  constructor(device: GPUDevice, timestampCount: number) {
+  constructor(device: GPUDevice, timestampPairCount: number) {
     this.timestampSupported = device.features.has('timestamp-query');
     if (!this.timestampSupported) return;
 
-    this.timestampCount = timestampCount;
+    this.timestamps = Array(timestampPairCount).fill(0);
 
     // Create timestamp queries
     this.timestampQuerySet = device.createQuerySet({
       type: 'timestamp',
-      count: timestampCount, // begin and end
+      count: timestampPairCount * 2, // begin and end
     });
 
     // Create a buffer where to store the result of GPU queries
     const timestampByteSize = 8; // timestamps are uint64
-    const timestampBufferSize = timestampCount * timestampByteSize;
     this.timestampBuffer = device.createBuffer({
-      size: timestampBufferSize,
+      size: this.timestampQuerySet.count * timestampByteSize,
       usage: GPUBufferUsage.COPY_SRC | GPUBufferUsage.QUERY_RESOLVE,
     });
 
     // Create a buffer to map the result back to the CPU
     this.timestampMapBuffer = device.createBuffer({
-      size: timestampBufferSize,
+      size: this.timestampBuffer.size,
       usage: GPUBufferUsage.COPY_DST | GPUBufferUsage.MAP_READ,
     });
+  }
 
-    this.hasOngoingTimestampReadback = false;
+  // Add both a start and end timestamp.
+  addTimestampWrite(
+    renderPassDescriptor: GPURenderPassDescriptor,
+    pairId: number
+  ) {
+    if (this.timestampSupported) {
+      // We instruct the render pass to write to the timestamp query before/after
+      const ndx = pairId * 2;
+      renderPassDescriptor.timestampWrites = {
+        querySet: this.timestampQuerySet,
+        beginningOfPassWriteIndex: ndx,
+        endOfPassWriteIndex: ndx + 1,
+      };
+    }
+    return renderPassDescriptor;
   }
 
   // Resolve all timestamp queries and copy the result into the map buffer
@@ -59,13 +70,13 @@ export default class TimestampQueryManager {
     commandEncoder.resolveQuerySet(
       this.timestampQuerySet,
       0 /* firstQuery */,
-      this.timestampCount /* queryCount */,
+      this.timestampQuerySet.count /* queryCount */,
       this.timestampBuffer,
       0 /* destinationOffset */
     );
 
-    if (!this.hasOngoingTimestampReadback) {
-      // Copy values to the mapped buffer
+    if (this.timestampMapBuffer.mapState === 'unmapped') {
+      // Copy values to the mappable buffer
       commandEncoder.copyBufferToBuffer(
         this.timestampBuffer,
         0,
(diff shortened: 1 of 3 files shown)


sample/timestampQuery/main.ts
```diff
@@ -222,32 +218,19 @@ function frame() {
...
-      'timestamps (ms): elapsed',
-      elapsedMs,
-      'avg',
-      renderPassDurationCounter.getAverage()
-    );
-    perfDisplay.innerHTML = `Render Pass duration: ${renderPassDurationCounter
-      .getAverage()
-      .toFixed(3)} ms ± ${renderPassDurationCounter.getStddev().toFixed(3)} ms`;
-  });
+  timestampQueryManager.update();
```
kainino0x, line 226: The update won't take effect until the map completes. This probably works fine because we'll just show the last data we _have_ gotten back, but putting `update()` here looks like it affects the code afterward.

If you want it to work like this you could maybe rename it to e.g. `tryInitiateTimestampDownload` and put it at the end next to rAF so it just looks like it's kicking off some async work for later, and comment here that `timestampQueryManager.timestamps` will just return the last data that's been downloaded.

sample/timestampQuery/TimestampQueryManager.ts
```diff
@@ -17,37 +14,51 @@ export default class TimestampQueryManager {
...
   timestampMapBuffer: GPUBuffer;
 
-  // State used to avoid firing concurrent readback of timestamp values
-  hasOngoingTimestampReadback: boolean;
+  // Last times
+  timestamps: number[];
 
   // Device must have the "timestamp-query" feature
-  constructor(device: GPUDevice, timestampCount: number) {
+  constructor(device: GPUDevice, timestampPairCount: number) {
```
kainino0x, line 21: if timestampPairCount is not actually used and in practice we only use indices 0/1, maybe remove this and the complexity that goes with it.

Also the call to `new TimestampQueryManager` still sets it to 2 but I think you're only using 1.

sample/timestampQuery/TimestampQueryManager.ts
```diff
@@ -17,37 +14,51 @@ export default class TimestampQueryManager {
   // A buffer to map this result back to CPU
   timestampMapBuffer: GPUBuffer;
 
-  // State used to avoid firing concurrent readback of timestamp values
-  hasOngoingTimestampReadback: boolean;
+  // Last times
+  timestamps: number[];
```
kainino0x, line 18: nit: name sounds like absolute timestamps, but actually contains time deltas

kainino0x (commented):
Few comments but I think this is good!
FYI @eliemichel

sample/timestampQuery/TimestampQueryManager.ts
greggman: renamed to `passesElaspedTimes` (hard to name).

kainino0x: maybe something like `passDurationMeasurements`

sample/timestampQuery/TimestampQueryManager.ts
greggman: Removed. Yes, it makes it simpler.

sample/timestampQuery/main.ts
greggman: done

beaufortfrancois:
It would be nice if compute boids sample is similar (either use timestampquery-manager or uses same variables): https://github.com/webgpu/webgpu-samples/blob/main/sample/computeBoids/main.ts

kainino0x (approved):

kainino0x:
> It would be nice if compute boids sample is similar (either use timestampquery-manager or uses same variables): https://github.com/webgpu/webgpu-samples/blob/main/sample/computeBoids/main.ts

Sharing the manager would be good! I'll land this but that can be a potential followup

eliemichel:
Thank you @greggman for the thorough feedback and retakes! I admit I am more used to playing with the C API rather than the JavaScript one, and some parts of the example were clearly imperfect.

Sorry I could not reply earlier, but I do not fully agree with the proposed change. I'll try my best to explain (but am off course open to counter arguments!):

> It's not clear what TimestampQueryManager was really managing.

The logic of the `TimestampQueryManager` is to be independent from the processing of the query results: the user says "My pipeline needs N timestamps, and here is what to do with the results (the callback)". I agree that my callback API was pretty bad indeed, it should have been a callback set once for all! But I still find it interesting to be a callback rather than moving the elapsed time computation logic inside the manager: if the user has multiple passes and wants to compute various timestamp difference, that's up to them.

You are effectively moving the behavior of `onTimestampReadBack` within the manager, making it specific, while otherwise the mechanism is generic (and the interpretation of timestamp indices is up to the user, hence it is manually specified in the pass descriptor (typically, using an enum if there are more than 2 timestamps).

I have been using a C equivalent of this manager for a while, and it is something I want to just copy-paste from one project to another without needing to customizing it all the time: it contains no domain-specific behavior.

**NB:** This is also why the query count was a user argument, and `resolveAll` called wuith "all".

> One other thing, I don't understand PerfCounter. If you get one bad outsized timing it seems like it will take an extremely long time to recover though maybe I don't understand the math.

If it is just for display it may not be ideal, but a typical use case is to measure perf for a benchmark report, which means giving an equal role to all samples. This just computes the mean/stddev w/o storing all samples.

Furthermore, a proper benchmarking requires to accurately know the number of samples. However, reading the stateful `passDurationMeasurementNs` rather than getting a callback leads to an incorrect sample count, because `addSample` is now called at each frame rather than each time a timestamp query gets read back (ofc one should also keep track of which passes were invoked in cases some of them do not run at each frame).

Of course a lot of this is mostly a question of genericity over minimalism and any outcome (your version, my proposition, or any other) would make as much sense in the end (the first version of the example was on the opposite side of the spectrum -- no abstraction at all -- and it was (rightfully) deemed hard to read). I only thinking that if, as suggested, we wanted to reuse this manager in other examples, it could be interesting to make it not-too-specific! Again, open to counter-arguments :)

PR #477, opened 2024-11-15
Fix perf display container background

beaufortfrancois:


greggman:
I don't think this is needed with #478 added?

PR #479, opened 2024-11-19
Blending example

greggman:
Here's the webgpufundamentals blending example ported to webgpu-samples.

Feel free to reject it if you think it's not a good example.

### DIFF
--- a/sample/blending/main.ts
+++ b/sample/blending/main.ts
@@ -1,0 +1,580 @@
+import { mat4 } from 'wgpu-matrix';
+import { GUI } from 'dat.gui';
+import { quitIfWebGPUNotAvailable } from '../util';
+import texturedQuadWGSL from './texturedQuad.wgsl';
+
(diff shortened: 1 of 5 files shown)


sample/blending/main.ts
```diff
@@ -0,0 +1,570 @@
...
+  'src-alpha-saturated',
+  'constant',
+  'one-minus-constant',
+];
+
+const presets: Record<
+  string,
+  { color?: GPUBlendComponent; alpha?: GPUBlendComponent }
+> = {
+  'default (copy)': {
```
kainino0x, line 276: The initial state of the page seems not to match what I get when I select this preset

kainino0x (approved):
This is great and super useful! Just one comment.

sample/blending/main.ts
greggman: fixed

PR #481, opened 2024-12-05
Address Comments for Timestamp example

greggman:
It was mentioned that a previous PR did the wrong thing by possibly applying a time more than once per timer query. This PR addresses that issue by making the TimestampQueryManager take a callback at initialization time that is called, only when a value is available. The value is provided to the callback. This pattern is similar to ResizeObserver

I agree with another comment, I kind of feel like it would be more useful to support multiple queries. But, maybe as a sample it's fine to support just 1.


kainino0x (approved):
LGTM

PR #488, opened 2025-01-17
Compatibility Mode, the easy part.

SenorBlanco:
Enable WebGPU Compatibility Mode for all samples that can run unmodified in that mode.

### DIFF
--- a/sample/animometer/main.ts
+++ b/sample/animometer/main.ts
@@ -3,7 +3,9 @@ import animometerWGSL from './animometer.wgsl';
 import { quitIfWebGPUNotAvailable } from '../util';
 
 const canvas = document.querySelector('canvas') as HTMLCanvasElement;
-const adapter = await navigator.gpu?.requestAdapter({ featureLevel: 'compatibility' });
+const adapter = await navigator.gpu?.requestAdapter({
+  featureLevel: 'compatibility'
+});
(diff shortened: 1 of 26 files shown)


toji:
Change looks fine, but I wanted to confirm the expected behavior:

 - For devices that *only* support compat mode this change will enable the sample to work where it didn't before.
 - For devices that support core WebGPU this flag will effectively be a no-op, even if the browser supports compat mode.
 - For browsers that do not implement compat mode this flag will be ignored.

Correct? If so, LGTM!

kainino0x:
> * For devices that support core WebGPU this flag will effectively be a no-op, even if the browser supports compat mode.

The sample will start getting using a compat (but d3d12/vulkana/metal) device in this case, but since the samples don't use any compat features it doesn't have an effect except for a little extra runtime validation

toji (approved):
Thanks for the clarification, Kai! LGTM, I just wanted to ensure this wasn't going to inadvertently downgrade the adapters for these samples.

kainino0x:
We realized this will break in Chrome <133 with `--enable-unsafe-webgpu`, and while that's not really a recommended configuration (people should be doing testing on Canary), it's going to be a bit before 133 stable rolls out, so reverting for now.

PR #489, opened 2025-01-21
Revert "Compatibility Mode, the easy part. (#488)"

SenorBlanco:
Revert "Compatibility Mode, the easy part. (#488)"

Chrome Stable with "Enable Unsafe WebGPU" enabled will currently error on these changes, since it requires that
the featureLevel string remain undefined. This will be fixed in the next stable release of Chrome, but it would be
unfortunate to break these samples in the meantime.

This reverts commit 0ce1ef2b244074fd631e85c490670fc0d19d2695.


kainino0x:
Oops, sorry I didn't realize you already had a PR open for this! I still had the tab open from earlier and it only took a few clicks so I did it already ^_^

SenorBlanco:
No worries! Thanks for jumping in.

PR #493, opened 2025-03-12
Compatibility Mode, the easy part.

SenorBlanco:
Enable WebGPU Compatibility Mode for all samples that can run unmodified in that mode.

### DIFF
--- a/sample/animometer/main.ts
+++ b/sample/animometer/main.ts
@@ -3,7 +3,9 @@ import animometerWGSL from './animometer.wgsl';
 import { quitIfWebGPUNotAvailable } from '../util';
 
 const canvas = document.querySelector('canvas') as HTMLCanvasElement;
-const adapter = await navigator.gpu?.requestAdapter();
+const adapter = await navigator.gpu?.requestAdapter({
+  featureLevel: 'compatibility',
+});
 const device = await adapter?.requestDevice();
 quitIfWebGPUNotAvailable(adapter, device);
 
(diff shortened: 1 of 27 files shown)


greggman (approved):

kainino0x (commented):
LGTM, but if we're waiting for Chrome 134, we should wait a bit longer because it doesn't ship on ChromeOS until next week. (https://chromiumdash.appspot.com/schedule)

greggman:
Committing now shouldn't break anything right? `featureLevel` will just be ignored.

SenorBlanco:
IIRC, it was fixed here: https://crrev.com/1392500.

I actually thought that went into M134, but according to this, it went into M133:

https://chromiumdash.appspot.com/commit/d00996faacd9c096f6c53617204b6ee5eee435c4

So we should be easily good to go now.

kainino0x:
great!

shrekshao:
I think with the new "core-features-and-limits" spec update, this pattern will create a compat device? (without requiring "core-features-and-limits", compat defaulting adapter will create compat device)

"core-features-and-limits"  is landed in M136 so we will need to update the samples by then.

Update will likely be:

```typescript
const adapter = await navigator.gpu?.requestAdapter({
  featureLevel: 'compatibility',
});
const device = await adapter?.requestDevice(adapter?.features.has('core-features-and-limits') ? {requiredFeatures: ['core-features-and-limits']} : {});
```

greggman:
It's okay to create a compat device as that's all the sample needs to run.

PR #494, opened 2025-03-13
Adjust wireframe example for compatibility mode

greggman:


sample/wireframe/main.ts
```diff
@@ -61,8 +61,17 @@ function createVertexAndIndexBuffer(
   };
 }
 
-const adapter = await navigator.gpu?.requestAdapter();
-const device = await adapter?.requestDevice();
+const adapter = await navigator.gpu?.requestAdapter({
+  featureLevel: 'compatibility',
+});
+if (adapter.limits.maxStorageBuffersInVertexStage < 2) {
```
kainino0x, line 67: This will be `undefined` (if not in chrome with compat enabled). Technically this would work because `undefined < 2` == `NaN < 2` == `false` but we should be somehow explicit about it.

I made the limit members optional in https://github.com/gpuweb/types/pull/168 which I think would catch this issue (but not the other one).

sample/wireframe/main.ts
```diff
@@ -61,8 +61,17 @@ function createVertexAndIndexBuffer(
...
+  fail('maxStorageBuffersInVertexStage is < 2');
+}
+const device = await adapter?.requestDevice({
+  requiredLimits: {
+    maxStorageBuffersInVertexStage: 2,
```
kainino0x, line 72: I believe this will produce an error in standard browsers. It should only be set if the limit exists in `adapter.limits`.

PR #495, opened 2025-03-13
Compat-ify samples (more difficult).

SenorBlanco:
These samples required changes beyond requesting compatibility mode.

a-buffer
* requires storage buffers in fragment stage (issue 19)
* depth texture binding cannot be used in `textureLoad()` (issue 16)
  * bind as `texture_2d<f32>` (and `unfilterable-float` `sampleType`) instead
* `interpolate(flat)` not supported (issue 17)
  * use `interpolate(flat, either)` instead

bitonicSort
* requires storage buffers in fragment stage (issue 19)

cubemap
* `textureBindingViewDimension` must be specified for cubemap sampling (issue 1)

deferredRendering
* depth texture binding cannot be used in `textureLoad()` (issue 16)
  * bind as `texture_2d<f32>` (and `unfilterable-float` `sampleType`) instead
* requires storage buffers in fragment stage (issue 19)

reversedZ
* depth texture binding cannot be used in `textureLoad()` (issue 16)
  * bind as `texture_2d<f32>` (and `unfilterable-float` `sampleType`) instead

samplerParameters
* requires storage buffers in vertex stage (issue 18)

skinnedMesh
* requires storage buffers in vertex stage (issue 18)

textRenderingMsdf
* requires storage buffers in vertex stage (issue 18)
* requires storage buffers in fragment stage (issue 19)

videoUploading
* no additional changes required (missed this one in the first pass)

wireframe
* requires storage buffers in vertex stage (issue 18)

### DIFF
--- a/sample/a-buffer/main.ts
+++ b/sample/a-buffer/main.ts
@@ -1,7 +1,7 @@
 import { mat4, vec3 } from 'wgpu-matrix';
 import { GUI } from 'dat.gui';
 
-import { quitIfWebGPUNotAvailable } from '../util';
+import { quitIfWebGPUNotAvailable, quitIfLimitLessThan } from '../util';
 import { mesh } from '../../meshes/teapot';
 
 import opaqueWGSL from './opaque.wgsl';
@@ -16,10 +16,8 @@ const canvas = document.querySelector('canvas') as HTMLCanvasElement;
 const adapter = await navigator.gpu?.requestAdapter({
   featureLevel: 'compatibility',
 });
-let limits: Record<string, GPUSize32>;
-if ('maxStorageBuffersInFragmentStage' in adapter.limits) {
-  limits = { maxStorageBuffersInFragmentStage: 2 };
-}
+const limits: Record<string, GPUSize32> = {};
+quitIfLimitLessThan(adapter, 'maxStorageBuffersInFragmentStage', 2, limits);
 const device = await adapter?.requestDevice({
   requiredLimits: limits,
 });
--- a/sample/a-buffer/main.ts
+++ b/sample/a-buffer/main.ts
@@ -13,8 +13,12 @@ function roundUp(n: number, k: number): number {
 }
 
 const canvas = document.querySelector('canvas') as HTMLCanvasElement;
-const adapter = await navigator.gpu?.requestAdapter();
-const device = await adapter?.requestDevice();
+const adapter = await navigator.gpu?.requestAdapter({
+  featureLevel: 'compatibility',
+});
+const device = await adapter?.requestDevice({
+  requiredLimits: { maxStorageBuffersInFragmentStage: 2 },
+});
 quitIfWebGPUNotAvailable(adapter, device);
 
 const context = canvas.getContext('webgpu') as GPUCanvasContext;
@@ -185,7 +189,7 @@ const translucentBindGroupLayout = device.createBindGroupLayout({
     {
       binding: 3,
       visibility: GPUShaderStage.FRAGMENT,
-      texture: { sampleType: 'depth' },
+      texture: { sampleType: 'unfilterable-float' },
     },
     {
       binding: 4,
--- a/sample/a-buffer/main.ts
+++ b/sample/a-buffer/main.ts
@@ -16,8 +16,12 @@ const canvas = document.querySelector('canvas') as HTMLCanvasElement;
 const adapter = await navigator.gpu?.requestAdapter({
   featureLevel: 'compatibility',
 });
+let limits: Record<string, GPUSize32>;
+if ('maxStorageBuffersInFragmentStage' in adapter.limits) {
+  limits = { maxStorageBuffersInFragmentStage: 2 };
+}
 const device = await adapter?.requestDevice({
-  requiredLimits: { maxStorageBuffersInFragmentStage: 2 },
+  requiredLimits: limits,
 });
 quitIfWebGPUNotAvailable(adapter, device);
 
--- a/sample/bitonicSort/utils.ts
+++ b/sample/bitonicSort/utils.ts
@@ -119,12 +119,16 @@ export const SampleInitFactoryWebGPU = async (
 
     const timestampQueryAvailable = adapter.features.has('timestamp-query');
     let features = [];
+    let limits: Record<string, GPUSize32>;
     if (timestampQueryAvailable) {
       features = ['timestamp-query'];
     }
+    if ('maxStorageBuffersInFragmentStage' in adapter.limits) {
+      limits = { maxStorageBuffersInFragmentStage: 1 };
+    }
     const device = await adapter.requestDevice({
       requiredFeatures: features,
-      requiredLimits: { maxStorageBuffersInFragmentStage: 1 },
+      requiredLimits: limits,
     });
     quitIfWebGPUNotAvailable(adapter, device);
 
--- a/sample/bitonicSort/utils.ts
+++ b/sample/bitonicSort/utils.ts
@@ -112,18 +112,20 @@ export const SampleInitFactoryWebGPU = async (
   callback: SampleInitCallback3D
 ): Promise<SampleInit> => {
   const init = async ({ canvas, gui, stats }) => {
-    const adapter = await navigator.gpu?.requestAdapter();
+    const adapter = await navigator.gpu?.requestAdapter({
+      featureLevel: 'compatibility',
+    });
     quitIfAdapterNotAvailable(adapter);
 
     const timestampQueryAvailable = adapter.features.has('timestamp-query');
-    let device: GPUDevice;
+    let features = [];
     if (timestampQueryAvailable) {
-      device = await adapter.requestDevice({
-        requiredFeatures: ['timestamp-query'],
-      });
-    } else {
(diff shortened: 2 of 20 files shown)


sample/a-buffer/main.ts
```diff
@@ -13,8 +13,12 @@ function roundUp(n: number, k: number): number {
...
 
 const canvas = document.querySelector('canvas') as HTMLCanvasElement;
-const adapter = await navigator.gpu?.requestAdapter();
-const device = await adapter?.requestDevice();
+const adapter = await navigator.gpu?.requestAdapter({
+  featureLevel: 'compatibility',
+});
+const device = await adapter?.requestDevice({
+  requiredLimits: { maxStorageBuffersInFragmentStage: 2 },
+});
```
greggman, line 21: @kainino0x pointed out this will fail on current chrome/firefox/safari

To make it work requires only adding it IF `maxStorageBuffersInFragmentStage` exists on `adapter.limits` because otherwise it's an unknown limit and will fail device creation.

There's a bunch of solutions - they all have trade offs.

Here's one

```suggestion
const adapter = await navigator.gpu?.requestAdapter({
  featureLevel: 'compatibility',
});
const { maxStorageBuffersInFragmentStage } = adapter.limits;

// Note: maxStorageBuffersInFragmentStage may be undefined in old browsers
// but that's ok as it will fail this check and old browser only support core
// which always has enough storage buffers in fragment shaders.
if (maxStorageBuffersInFragmentStage < 2) {
  fail('maxStorageBuffersInFragmentStage < 2'); // export from util.ts
}
const device = await adapter?.requestDevice({
  requiredLimits: { maxStorageBuffersInFragmentStage },
);
```

sample/bitonicSort/utils.ts
```diff
@@ -112,18 +112,20 @@ export const SampleInitFactoryWebGPU = async (
...
-      device = await adapter.requestDevice({
-        requiredFeatures: ['timestamp-query'],
-      });
-    } else {
-      device = await adapter.requestDevice();
+      features = ['timestamp-query'];
     }
+    const device = await adapter.requestDevice({
+      requiredFeatures: features,
+      requiredLimits: { maxStorageBuffersInFragmentStage: 1 },
```
greggman, line 127: same as above

sample/deferredRendering/main.ts
```diff
@@ -15,8 +15,12 @@ const lightExtentMin = vec3.fromValues(-50, -30, -50);
 const lightExtentMax = vec3.fromValues(50, 50, 50);
 
 const canvas = document.querySelector('canvas') as HTMLCanvasElement;
-const adapter = await navigator.gpu?.requestAdapter();
-const device = await adapter?.requestDevice();
+const adapter = await navigator.gpu?.requestAdapter({
+  featureLevel: 'compatibility',
+});
+const device = await adapter?.requestDevice({
+  requiredLimits: { maxStorageBuffersInFragmentStage: 1 },
```
greggman, line 22: same as above

sample/samplerParameters/main.ts
```diff
@@ -28,8 +28,12 @@ const kMatrices: Readonly<Float32Array> = new Float32Array([
 ]);
 
 const canvas = document.querySelector('canvas') as HTMLCanvasElement;
-const adapter = await navigator.gpu?.requestAdapter();
-const device = await adapter?.requestDevice();
+const adapter = await navigator.gpu?.requestAdapter({
+  featureLevel: 'compatibility',
+});
+const device = await adapter?.requestDevice({
+  requiredLimits: { maxStorageBuffersInVertexStage: 1 },
```
greggman, line 35: same as above

sample/skinnedMesh/main.ts
```diff
@@ -95,8 +95,12 @@ const getRotation = (mat: Mat4): Quat => {
 
 //Normal setup
 const canvas = document.querySelector('canvas') as HTMLCanvasElement;
-const adapter = await navigator.gpu?.requestAdapter();
-const device = await adapter?.requestDevice();
+const adapter = await navigator.gpu?.requestAdapter({
+  featureLevel: 'compatibility',
+});
+const device = await adapter?.requestDevice({
+  requiredLimits: { maxStorageBuffersInVertexStage: 2 },
```
greggman, line 102: same as above

sample/textRenderingMsdf/main.ts
```diff
@@ -14,8 +14,15 @@ import vertexPositionColorWGSL from '../../shaders/vertexPositionColor.frag.wgsl
...
 const canvas = document.querySelector('canvas') as HTMLCanvasElement;
-const adapter = await navigator.gpu?.requestAdapter();
-const device = await adapter?.requestDevice();
+const adapter = await navigator.gpu?.requestAdapter({
+  featureLevel: 'compatibility',
+});
+const device = await adapter?.requestDevice({
+  requiredLimits: {
+    maxStorageBuffersInFragmentStage: 1,
+    maxStorageBuffersInVertexStage: 2,
```
greggman, line 23: same as above

greggman (changes requested):
See comments. The pattern in the PR would be okay in say 6 months but it doesn't work today.

sample/a-buffer/main.ts
greggman: Note: I'm not sure that example is acceptable nor not since it's not requesting 2, it's requesting max. The more I have to deal with features and limits the more think you should just always get max limits. It's much easier to deal with.

```js
const adapter = await navigator.gpu?.requestAdapter({
  featureLevel: 'compatibility',
});
const device = await adapter.requestDevice({
  requiredFeatures: adapter.features,
  requiredLimits: objLikeToObj(adapter.limits), // need a helper ATM
});
if (device.limits.maxStorageBuffersInFragmentStage < 2) {
  fail('I can has no running');
}
```

It's way simpler than the summersault of checking for in adapter and applying in device but it's not the intent of the spec.

SenorBlanco: Thanks.

> const { maxStorageBuffersInFragmentStage } = adapter.limits;

gave me:

```
(!) [plugin typescript] sample/a-buffer/main.ts (20:9): @rollup/plugin-typescript TS2339: Property 'maxStorageBuffersInFragmentStage' does not exist on type 'GPUSupportedLimits'.
/home/senorblanco/src/webgpu-samples/sample/a-buffer/main.ts:20:9
```

I used 

```
if ('maxStorageBuffersInFragmentStage' in adapter.limits) {
```

instead. LMK if that's kosher.

SenorBlanco: > Note: I'm not sure that example is acceptable nor not since it's not requesting 2, it's requesting max. The more I have to deal with features and limits the more think you should just always get max limits. It's much easier to deal with.

It is easier, but I think it's dangerous, in that it turns WebGPU into OpenGL: you could write some code that runs fine on your machine (with above-average limits), but won't run elsewhere (with default limits). I'd prefer to request the minimum that the sample actually needs.

greggman: > const { maxStorageBuffersInFragmentStage } = adapter.limits;

It gets an error because the types need to be updated

```
npm install --save-dev @webgpu/types@latest
```

sample/a-buffer/main.ts
```diff
@@ -13,8 +13,16 @@ function roundUp(n: number, k: number): number {
...
-const adapter = await navigator.gpu?.requestAdapter();
-const device = await adapter?.requestDevice();
+const adapter = await navigator.gpu?.requestAdapter({
+  featureLevel: 'compatibility',
+});
+let limits: Record<string, GPUSize32>;
+if ('maxStorageBuffersInFragmentStage' in adapter.limits) {
+  limits = { maxStorageBuffersInFragmentStage: 2 };
+}
+const device = await adapter?.requestDevice({
```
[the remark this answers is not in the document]
SenorBlanco: Good catch, thanks. I thought the samples were checking for requestDevice() failure, but I see they're not.

I've implemented better limit checking. LMK what you think.

sample/util.ts
```diff
@@ -11,6 +11,24 @@ export function quitIfAdapterNotAvailable(
...
+export function quitIfLimitNotMet(
+  adapter: GPUAdapter,
+  limit: string,
+  requiredValue: number,
+  limits: Record<string, GPUSize32>
+) {
+  if (limit in adapter.limits) {
+    const limitKey = limit as keyof GPUSupportedLimits;
+    const limitValue = adapter.limits[limitKey] as number;
+    if (limitValue < requiredValue) {
```
greggman, line 23: nit: some limits are `minXXX` instead of `maxXXX`. Up to you if you want if it now or wait for someone to stumble on it.

greggman (approved):
lgtm! one nit.

sample/a-buffer/main.ts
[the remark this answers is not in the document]
greggman: `requestDevice` never fails (never turns null/undefined) so "if called correctly" there's nothing to check

sample/util.ts
SenorBlanco: Renamed QuitIfLimitNotMet() -> QuitIfLimitLessThan(). Someone else can make the GreaterThan flavour if desired.

PR #497, opened 2025-03-21
Add generate mipmap sample

greggman:
We've got hello triangle. Generating mipmaps seems like a pretty common feature. Maybe a sample for it is good?

### DIFF
--- a/sample/generateMipmap/generateMipmap.ts
+++ b/sample/generateMipmap/generateMipmap.ts
@@ -1,0 +1,164 @@
+import generateMipmapWGSL from './generateMipmap.wgsl';
+
+export function numMipLevels(...sizes: number[]) {
+  const maxSize = Math.max(...sizes);
+  return (1 + Math.log2(maxSize)) | 0;
+}
+
+/**
+ * Get the default viewDimension
+ * Note: It's only a guess. The user needs to tell us to be
+ * correct in all cases because we can't distinguish between
+ * a 2d texture and a 2d-array texture with 1 layer, nor can
+ * we distinguish between a 2d-array texture with 6 layers and
+ * a cubemap.
+ */
+export function getDefaultViewDimensionForTexture(
+  dimension: GPUTextureDimension,
+  depthOrArrayLayers: number
+) {
+  switch (dimension) {
+    case '1d':
+      return '1d';
+    default:
+    case '2d':
+      return depthOrArrayLayers > 1 ? '2d-array' : '2d';
+    case '3d':
+      return '3d';
+  }
+}
+
+type DeviceSpecificInfo = {
+  sampler: GPUSampler;
+  module: GPUShaderModule;
+  pipelineByFormatAndView: Map<string, GPURenderPipeline>;
+};
+
+function createDeviceSpecificInfo(device: GPUDevice): DeviceSpecificInfo {
+  const module = device.createShaderModule({
+    label: 'textured quad shaders for mip level generation',
+    code: generateMipmapWGSL,
+  });
+
+  const sampler = device.createSampler({
+    minFilter: 'linear',
+    magFilter: 'linear',
+  });
+
+  return {
+    module,
+    sampler,
+    pipelineByFormatAndView: new Map(),
+  };
+}
+
+const s_deviceToDeviceSpecificInfo = new WeakMap<
+  GPUDevice,
+  DeviceSpecificInfo
+>();
+
+/**
+ * Generates mip levels 1 to texture.mipLevelCount - 1 using
+ * mip level 0 as the source.
+ * @param device The device
+ * @param texture The texture to generate mips for
+ * @param textureBindingViewDimension The view dimension, needed for
+ *   compatibility mode if the texture is a cube map OR if the texture
+ *   is a 1 layer 2d-array.
+ */
+export function generateMips(
+  device: GPUDevice,
+  texture: GPUTexture,
+  textureBindingViewDimension?: GPUTextureViewDimension
+) {
+  textureBindingViewDimension =
+    textureBindingViewDimension ??
+    getDefaultViewDimensionForTexture(
+      texture.dimension,
+      texture.depthOrArrayLayers
+    );
+  const deviceSpecificInfo =
+    s_deviceToDeviceSpecificInfo.get(device) ??
+    createDeviceSpecificInfo(device);
+  s_deviceToDeviceSpecificInfo.set(device, deviceSpecificInfo);
+  const { sampler, module, pipelineByFormatAndView } = deviceSpecificInfo;
+
+  const id = `${texture.format}.${textureBindingViewDimension}`;
+
+  if (!pipelineByFormatAndView.get(id)) {
+    // Choose an fragment shader based on the viewDimension (removes the '-' from 2d-array and cube-array)
+    const entryPoint = `fs${textureBindingViewDimension.replace('-', '')}`;
+    pipelineByFormatAndView.set(
+      id,
+      device.createRenderPipeline({
+        label: `mip level generator pipeline for ${textureBindingViewDimension}, format: ${texture.format}`,
+        layout: 'auto',
+        vertex: {
+          module,
+        },
+        fragment: {
+          module,
+          entryPoint,
+          targets: [{ format: texture.format }],
+        },
+      })
+    );
+  }
+
+  const pipeline = pipelineByFormatAndView.get(id);
+  const encoder = device.createCommandEncoder({
+    label: 'mip gen encoder',
+  });
+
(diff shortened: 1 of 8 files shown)


sample/generateMipmap/main.ts
```diff
@@ -0,0 +1,113 @@
...
+
+for (let baseMipLevel = 0; baseMipLevel < mipLevelCount; ++baseMipLevel) {
+  const width = Math.max(1, imageTexture.width >> baseMipLevel);
+  const height = Math.max(1, imageTexture.height >> baseMipLevel);
+  canvas.width = width;
+  canvas.height = height;
+
+  const commandEncoder = device.createCommandEncoder();
+  const t = context.getCurrentTexture();
+  console.log(t.width, t.height);
```
beaufortfrancois, line 63: Please remove this leftover

sample/generateMipmap/main.ts
```diff
@@ -0,0 +1,113 @@
...
+
+  passEncoder.setBindGroup(0, bindGroup);
+  passEncoder.draw(6);
+  passEncoder.end();
+
+  device.queue.submit([commandEncoder.finish()]);
+
+  canvas.toBlob((blob) => {
+    const img = new Image();
+    img.src = URL.createObjectURL(blob);
```
beaufortfrancois, line 102: Shall we call revokeObjectURL when img is loaded?

sample/generateMipmap/main.ts
```diff
@@ -0,0 +1,113 @@
...
+    document.body.append(img);
+  });
+}
+
+const settings = { expand: false };
```
beaufortfrancois, line 107: FWIW the word "expand" was confusing to me until I tried it.

sample/generateMipmap/main.ts
```diff
@@ -0,0 +1,113 @@
...
+const gui = new GUI();
+gui.add(settings, 'expand').onChange((expand) => {
+  document.querySelectorAll('img').forEach((img) => {
```
beaufortfrancois, line 110: How about simply adding this class to document.body instead of looping and update CSS to `body.expand img { width: 512px... }`?

sample/generateMipmap/main.ts
```diff
@@ -0,0 +1,113 @@
...
+    ],
+  });
+
+  passEncoder.setBindGroup(0, bindGroup);
+  passEncoder.draw(6);
+  passEncoder.end();
+
+  device.queue.submit([commandEncoder.finish()]);
+
+  canvas.toBlob((blob) => {
```
beaufortfrancois, line 100: The only thing with that pattern is that the order is not necessarily the expected one (from large to small):
localhost_8080__sample=generateMipmap(Nexus 5X)

beaufortfrancois (commented):
Here's my drive-by review! Thanks @greggman for documenting those!

sample/generateMipmap/main.ts
greggman: done

sample/generateMipmap/main.ts
greggman: dome

sample/generateMipmap/generateMipmap.ts
```diff
@@ -0,0 +1,164 @@
...
+ * @param device The device
+ * @param texture The texture to generate mips for
+ * @param textureBindingViewDimension The view dimension, needed for
+ *   compatibility mode if the texture is a cube map OR if the texture
+ *   is a 1 layer 2d-array.
+ */
+export function generateMips(
+  device: GPUDevice,
+  texture: GPUTexture,
+  textureBindingViewDimension?: GPUTextureViewDimension
```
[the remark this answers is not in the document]
greggman: That's the opposite of your advice when I wrote tests in the past. You've said in the past, only set `textureBindingViewDimension` if it's required, otherwise rely on the defaults.

I think someone coming into this just wants it to work without having the specify anything. It will work except for cube and 1 layer 2d-array. No one is likely to make a 1-layer 2d array so that leaves cube. For cube they'll get a validation error alone the lines of `tried to use a 2d-array as a cube, can't do that in compat mode`

sample/generateMipmap/main.ts
greggman: why? It just clutters the sample.

```
img.decode().then(() => URL.revokeObjectURL(img.src))
```

it also makes the image not viewable if you right click and pick "Open Image in New Tab". Similarly if you click the link in DevTools.

beaufortfrancois: It's best practice to revokeObjectURL when image is not needed anymore. The "Shall" meant it was a question ;)

sample/generateMipmap/main.ts
greggman: fixed

sample/generateMipmap/generateMipmap.ts
```diff
@@ -0,0 +1,164 @@
+import generateMipmapWGSL from './generateMipmap.wgsl';
+
+export function numMipLevels(...sizes: number[]) {
+  const maxSize = Math.max(...sizes);
+  return (1 + Math.log2(maxSize)) | 0;
+}
+
+/**
+ * Get the default viewDimension
+ * Note: It's only a guess. The user needs to tell us to be
```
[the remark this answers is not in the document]
greggman: The sample works for cube maps. The reason it doesn't pass the `textureBindingViewDimension` from the `GPUTextureDescriptor` is because the descriptor is unavailable. `generateMips` only input is the `GPUTexture`. The `textureBindingViewDimesion` is lost after texture creation. It's common to call `generateMips` after updating the contents of the first mip and it would be inconvenient if your code that wasn't keeping the texture descriptors around suddenly had to keep them around just for this function. As it is you don't have to keep the descriptor around. You just pass `'cube'` in the case of cubemaps, otherwise it just works.

sample/generateMipmap/main.ts
greggman: changed to `view 512x512`. not sure what else to change it to `dat.gui` defaults to only around 12 characters.

sample/generateMipmap/generateMipmap.ts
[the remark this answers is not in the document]
greggman: I could add examples of all 4 types of textures 2d, 2d-array, cube, cube-array (skipped in compat). It would make the sample larger. Then the question of how to show them. I guess I could spinning draw cubes with and without mipmaps

sample/generateMipmap/generateMipmap.ts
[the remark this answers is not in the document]
SenorBlanco: Yes, I agree with myself. :) Don't pass tbvd if you don't need to. It's just that it seems a bit sad to have to duplicate the browser's default logic for it, but perhaps that's unavoidable if you want to have the simple case be simple (2D).

What I'm getting at is, the sample doesn't show how to correctly generate cube maps, even though the utility function suggests it can do it. So I feel like the utility function should either be stripped down to just 2d, or the sample should exercise cube maps as well.

All that said, I think this is a great sample and it's also fine as is.

sample/generateMipmap/generateMipmap.ts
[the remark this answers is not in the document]
kainino0x: If `textureBindingViewDimension` were reflected from the GPUTexture, could we get rid of this helper?
We already reflect [other stuff](https://gpuweb.github.io/gpuweb/#gputexture), it seems like we can reflect this too, especially if it's useful.

In Core we would probably return `undefined`.

kainino0x (approved):
No additional comments from me!

sample/generateMipmap/generateMipmap.ts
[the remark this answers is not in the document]
greggman: I updated to show 4 cubes. One uses 2d, one 2d-array, one cube, one cube-array if it can, else cube

SenorBlanco:
Hmmm.. on Chrome Linux Stable (134), I get:

Uncaptured error:
Dimension (TextureViewDimension::Cube) of [TextureView of Texture (unlabeled 256x256 px, 6 layer, TextureFormat::RGBA8Unorm)] must match textureBindingViewDimension (TextureViewDimension::e2DArray) of [Texture (unlabeled 256x256 px, 6 layer, TextureFormat::RGBA8Unorm)] in compatibility mode.
 - While validating entries[2] as a Sampled Texture.
Expected entry layout: {sampleType: TextureSampleType::Float, viewDimension: 4, multisampled: 0}
 - While validating [BindGroupDescriptor] against [BindGroupLayout (unlabeled)]
 - While calling [Device].CreateBindGroup([BindGroupDescriptor]).

greggman:
Sorry, it should be fixed now

PR #501, opened 2025-04-17
Add bc4-r-unorm and astc-12x12-unorm brain images

beaufortfrancois:
This PR adds the possibility for users to pick bc4-r-unorm and astc-12x12-unorm brain images in the "Volume Rendering - Texture 3D" sample.


beaufortfrancois:
FYI @mehmetoguzderin

beaufortfrancois:
@mwyrzykowski I've noticed using the `bc4-r-unorm` texture format in this sample is causing the canvas camera to be stuck somehow... Is this a known issue with Safari Tech Preview or am I holding it wrong?


https://github.com/user-attachments/assets/482997ea-73e6-4c9b-9679-d233a20b0fb4

sample/volumeRenderingTexture3D/main.ts
```diff
@@ -7,21 +7,51 @@ const canvas = document.querySelector('canvas') as HTMLCanvasElement;
...
   near: 2.0,
   far: 7.0,
+  textureFormat: 'r8unorm',
 };
 
 gui.add(params, 'rotateCamera', true);
 gui.add(params, 'near', 2.0, 7.0);
 gui.add(params, 'far', 2.0, 7.0);
+gui.add(params, 'textureFormat', Object.keys(brainImages)).onChange(() => {
+  createVolumeTexture(params.textureFormat);
```
kainino0x, line 42: On Chrome Stable (and presumably on Canary on hardware that doesn't support BC) I get this when I switch:

> Uncaught (in promise) TypeError: Failed to execute 'createTexture' on 'GPUDevice': Use of the 'bc4-r-unorm' texture format requires the 'texture-compression-bc' feature to be enabled on [Device].

This sample should do something to gracefully fall back (at least, don't throw an error and then keep showing the r8unorm image).

Also, maybe the Safari issue is that it's failing silently instead of throwing an exception?

mwyrzykowski:
Ah looks like we were rejecting 3D compressed texture creation even when the extension was enabled, thanks for catching @beaufortfrancois 

I suppose the sample could detect the validation error we generated but I addressed it with https://bugs.webkit.org/show_bug.cgi?id=291690:

https://github.com/user-attachments/assets/fd54e05d-48e5-45d6-b2b8-fb4fd3927167

sample/volumeRenderingTexture3D/main.ts
beaufortfrancois: Instead of `Object.keys(brainImages)`, we could use `device.features` to selectively add only the texture formats that we know will work. However, I've kept all of them initially so that users can see that the lack of device features is why it's not working: WebGPU supports BC and ASTC but not this machine.

kainino0x: That's fine, I do think we should list all of the formats supported by the sample even on systems that don't support them. The problem is if I select an unsupported format (like BC on Chrome Stable right now) then it silently fails to update the texture, then keeps rendering with the r8unorm version. There's no indication that it failed unless I open dev tools.

beaufortfrancois: I'm surprised `device.onuncapturederror` is not triggered.
https://github.com/webgpu/webgpu-samples/blob/85e141cd46bad0e69982f2baab1bc4f4b6eccb22/sample/util.ts#L49-L51

I'll have a look tomorrow.

beaufortfrancois: Oh that's because `createTexture` does not generate a validation error, but a typeerror! I'll fix it.

beaufortfrancois: That's much better indeed! Thank you @kainino0x

beaufortfrancois: FYI next to come is https://github.com/beaufortfrancois/webgpu-samples/compare/bc4...beaufortfrancois:webgpu-samples:astc12?expand=1

sample/volumeRenderingTexture3D/main.ts
```diff
@@ -7,21 +7,51 @@ const canvas = document.querySelector('canvas') as HTMLCanvasElement;
...
 const gui = new GUI();
 
+const brainImages = {
+  r8unorm: {
+    bytesPerBlock: 1,
+    blockLength: 1,
+    dataPath:
+      '../../assets/img/volume/t1_icbm_normal_1mm_pn0_rf0_180x216x180_uint8_1x1.bin-gz',
+  },
+  'bc4-r-unorm': {
```
kainino0x, line 17: One more thought, could you put a comment here about what tool you used to compress the image?

mwyrzykowski:
> > Ah looks like we were rejecting 3D compressed texture creation even when the extension was enabled, thanks for catching
> 
> @mwyrzykowski Is it firing the GPUDevice `uncapturederror` event? The samples are supposed to pop up a dialog if that happens.

@kainino0x it appears the related CTS tests https://gpuweb.github.io/cts/standalone/?q=webgpu:api,validation,error_scope:* and https://gpuweb.github.io/cts/standalone/?q=webgpu:api,operation,uncapturederror:* are passing in WebKit as the pattern from https://www.w3.org/TR/webgpu/#example-bf3af559 works in WebKit

But the method used in the samples:
```
  device.onuncapturederror = (ev) => {
    fail(`Uncaptured error:\n${ev.error.message}`);
  };
```
does not. Replacing that with:
```
    device.addEventListener('uncapturederror', (ev) => {
        fail(`Uncaptured error:\n${ev.error.message}`);
    });
```
results in:
Screenshot 2025-04-17 at 5 14 42 PM

sample/volumeRenderingTexture3D/main.ts
mehmetoguzderin: Thank you very much once again for the realization of this feature, it is exciting to see it come to the stage of going live. For preparation of BC4 version, I used following script after first converting the 1x1 binary version to PNG slices. It incorporates wine since I just wanted to download texconv from GitHub releases and run on macOS (executes more than well). Your consideration would be appreciated!

```python
import os
import subprocess

# Update these paths according to your setup
source_directory = "./slices"
target_directory = "./compress_slices"
texconv_path = "./texconv.exe"
final_file_path = "./compress/t1_icbm_normal_1mm_pn0_rf0_180x216x180_bc4_4x4.bin"

# Ensure the output directory exists
os.makedirs(target_directory, exist_ok=True)

# Convert PNG files to BC4 format
for file_name in os.listdir(source_directory):
    if file_name.lower().endswith(".png"):
        source_path = os.path.join(source_directory, file_name)
        target_file_name = file_name[:-4] + ".dds"  # Change extension from .png to .dds
        target_path = os.path.join(target_directory, target_file_name)

        # Run texconv with Wine to convert the image
        subprocess.run(
            [
                "wine",
                texconv_path,
                "-f",
                "BC4_UNORM",
                "-y",
                "-m",
                "1",
                "-o",
                target_directory,
                "-ft",
                "dds",
                source_path,
            ],
            check=True,
        )

# Concatenate BC4 files into a single binary file, excluding headers
with open(final_file_path, "wb") as output_file:
    for file_name in sorted(os.listdir(target_directory)):  # Sort if order matters
        if file_name.lower().endswith(".dds"):
            path = os.path.join(target_directory, file_name)

            # Read and append the file content, skipping the header
            with open(path, "rb") as input_file:
                # Skip DDS header, typically 128 bytes. Adjust if different.
                content = input_file.read()[128:]
                output_file.write(content)

print("Conversion and concatenation complete.")
```

beaufortfrancois:
@mwyrzykowski My understanding is that both `device.onuncapturederror = ...` and `device.addEventListener('uncapturederror', ...` should work. Is this an STP issue?

kainino0x:
Thank you for investigating, Mike! Filed CTS bug https://github.com/gpuweb/cts/issues/4361

sample/volumeRenderingTexture3D/main.ts
kainino0x: Thanks for the reference!

For the comment, I think just saying that it was generated using Microsoft's Texconv is enough info.

mehmetoguzderin: Agreed!

kainino0x: Actually it seems there are two different incarnations of Texconv so a link would be useful.

mehmetoguzderin: I used an older version but from here: https://github.com/microsoft/DirectXTex/releases/tag/mar2025

mehmetoguzderin, line 17:
```suggestion
  'bc4-r-unorm': { // Generated with texconv from https://github.com/microsoft/DirectXTex/releases
```

sample/volumeRenderingTexture3D/main.ts
beaufortfrancois: I've updated the code to stop animation when downloading images to make things clearer.

beaufortfrancois:
To illustrate the value of both formats, here's a Chromium DevTools network screenshot:

woohoo

mwyrzykowski:
> @mwyrzykowski My understanding is that both `device.onuncapturederror = ...` and `device.addEventListener('uncapturederror', ...` should work. Is this an STP issue?

Yes, presumably so

mwyrzykowski (approved):

beaufortfrancois:
> > @mwyrzykowski My understanding is that both `device.onuncapturederror = ...` and `device.addEventListener('uncapturederror', ...` should work. Is this an STP issue?
> 
> Yes, presumably so

@mwyrzykowski I've tried https://red-triangle-webgpu.glitch.me/uncapturederror.html in Chrome, Firefox Nightly, and Safari Tech Preview.
Both Chrome and Firefox Nightly log `"onuncapturederror"` and `"addEventListener("uncapturederror")"` while Safari Tech Preview only logs `"addEventListener("uncapturederror")"`

sample/volumeRenderingTexture3D/main.ts
kainino0x: I am still seeing an unhandled promise rejection when I try this in Chrome Stable. I don't know what we should show, but not that, because it says "unhandled promise rejection, please report a bug!"

sample/volumeRenderingTexture3D/main.ts
```diff
@@ -7,21 +7,71 @@ const canvas = document.querySelector('canvas') as HTMLCanvasElement;
...
+  textureFormat: 'r8unorm',
 };
 
 gui.add(params, 'rotateCamera', true);
 gui.add(params, 'near', 2.0, 7.0);
 gui.add(params, 'far', 2.0, 7.0);
+gui
+  .add(params, 'textureFormat', Object.keys(brainImages))
+  .onChange(async () => {
+    cancelAnimationFrame(rafId);
```
kainino0x: nit: clear the screen instead of just stopping animation, in case animation is paused already, so it's clearer

sample/volumeRenderingTexture3D/main.ts
kainino0x: Pushed a commit to fix this

sample/volumeRenderingTexture3D/main.ts
```diff
@@ -4,24 +4,76 @@ import volumeWGSL from './volume.wgsl';
...
 
 const gui = new GUI();
 
+const brainImages = {
+  r8unorm: {
+    bytesPerBlock: 1,
+    blockLength: 1,
+    feature: undefined,
+    dataPath:
+      '../../assets/img/volume/t1_icbm_normal_1mm_pn0_rf0_180x216x180_uint8_1x1.bin.gz',
```
kainino0x, line 17: future improvement: brotli would probably be better than gzip

kainino0x (approved):

sample/volumeRenderingTexture3D/main.ts
kainino0x: update: Since we realized we need to use DecompressionStream, we are stuck with gzip (or deflate).

beaufortfrancois:
FYI https://github.com/whatwg/compression/issues/34 shows that chrome won't
have brotli support soon sadly.

kainino0x:
The concerns there seem to be about brotli compression, but we only need decompression.

beaufortfrancois:
According to https://github.com/whatwg/compression/issues/34#issuecomment-2824592203, Safari has Brotli support to Compression Streams by using Apple's Compression Framework. See https://github.com/WebKit/WebKit/commit/0d3451c531c810977123d2b9503ed968381d2004

mdanowar3 (approved):

Issue #504, opened 2025-04-18
volumeRenderingTexture3D bug?


greggman:
When I go to: https://webgpu.github.io/webgpu-samples/?sample=volumeRenderingTexture3D

I get

> Uncaptured error:
Required size for texture data layout (6998400) exceeds the linear data size (3788907) with offset (0).
> - While calling [Queue].WriteTexture([TexelCopyTextureInfo texture: [Texture (unlabeled 180x216x180 px, TextureFormat::R8Unorm)], mipLevel: 0, origin: [Origin3D x:0, y:0, z:0], aspect: TextureAspect::All], (3788907 bytes), [TexelCopyBufferLayout offset:0, bytesPerRow:180, rowsPerImage:216], [Extent3D width:180, height:216, depthOrArrayLayers:180])

greggman:
Strange, I don't get this when running from localhost on top of tree

greggman:
Just guessing it's because these files end in `.gz` and are treated differently by the github pages server vs the dev server.?

greggman:
Yea, when served from the dev server

```
content-encoding: gzip
content-length: 3788907
content-type: application/octet-stream
```

when served from github pages its

```
content-length: 3788907
content-type: application/gzip
```

content-encoding so the browser doesn't auto-un-gzip

github pages will auto compress certain files types. We probably shouldn't be using .gz and just include the uncompressed versions and let the server decide whether or not to serve them compressed? 

Should I turn off the content-type on the dev server?

greggman:
related: https://github.com/orgs/community/discussions/21655

greggman:
It's not clear what extension to use `.bin` files are not auto-compressed by github pages.  `.glb` are `.wasm` are. There is apparently no official list from github of the extensions that are served compressed.

If we just want the sample to work we can just uncompress them as their .bin files

If you want to download and decompress them yourself there's [unzipit](https://greggman.github.io/unzipit/) which it wrote. It's minimal, low-dependency, and has a simple interface

kainino0x:
This was working before #501, and we were already using gzip for the non-gpu-compressed data. Was this caused by the switch from `.bin-gz` to `.bin.gz`?

kainino0x:
We can always do a ServiceWorker hack, e.g. https://github.com/jprendes/service-worker-brotli

beaufortfrancois:
[...]

PR #506, opened 2025-04-19
Fix volume rendering texture 3d sample

greggman:
The files were changed from `.bin-gz` to `.bin.gz` and then left
to the browser to decompress. But, that only happened to work
because the local dev server served `.gz` files with
`content-encoding: gzip` headers. Github doesn't do this.

Revert to using `.bin-gz` and manually decompressing with
`DecompressionStream`. `.bin-gz` seems better then `.gz` because
many servers are configured to treat files that end in `.gz` specially.

* #504


mehmetoguzderin:
Just for background, when initially contributing the sample, I went with gz to also demonstrate how compression formats interact with archival compression such as gzip or brotli, and also out of concern for git repo size (hence use of bin-gz as extension to explicitly control when things decompress). However either way should be fine

beaufortfrancois (commented):
The `public/assets/img/volume/t1_icbm_normal_1mm_pn0_rf0.py` file would also be need to be updated right as it generates a gzip file. See `gzip_filename` in https://github.com/webgpu/webgpu-samples/pull/501/commits/af6432c6b556a1a2abdb6e17c54f66f39756ff2f

greggman:
updated

sample/volumeRenderingTexture3D/main.ts
```diff
@@ -158,11 +158,20 @@ async function createVolumeTexture(format: GPUTextureFormat) {
   });
 
   const response = await fetch(dataPath);
-  const buffer = await response.arrayBuffer();
-
+  const compressedBlob = await response.blob();
```
beaufortfrancois, line 161: Out of curiosity, why blob?

beaufortfrancois (approved):

sample/volumeRenderingTexture3D/main.ts
greggman: because theoretically it uses less memory? The browser can stream the blob through the decompressor. It can't stream the arrayBuffer as it's required to have loaded the entire thing

beaufortfrancois: Gotcha! Thanks!

PR #507, opened 2025-04-19
Make `fetch`/`DecompressionStream` usage more canonical.

greggman:
The `body` property of a `Response` from `fetch` is a stream.


beaufortfrancois (approved):
Much better indeed!

PR #510, opened 2025-04-21
Reland "Show error dialog on unhandled rejection and uncaught exception"

kainino0x:
Forgot this util is also used on workers. Reland with fix ~(and unrelated minor nit fixes)~.

Original landed in #502, reverted in #509.

### DIFF
--- a/sample/util.ts
+++ b/sample/util.ts
@@ -1,3 +1,14 @@
+// Show an error dialog if there's any uncaught exception or promise rejection.
+// This gets set up on all pages that include util.ts.
+globalThis.addEventListener('unhandledrejection', (ev) => {
+  fail(`unhandled promise rejection, please report a bug!
+  https://github.com/webgpu/webgpu-samples/issues/new\n${ev.reason}`);
+});
+globalThis.addEventListener('error', (ev) => {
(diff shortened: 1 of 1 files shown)


sample/util.ts
```diff
@@ -40,19 +51,18 @@ export function quitIfWebGPUNotAvailable(
...
     quitIfAdapterNotAvailable(adapter);
     fail('Unable to get a device for an unknown reason');
-    return;
   }
 
   device.lost.then((reason) => {
     fail(`Device lost ("${reason.reason}"):\n${reason.message}`);
   });
-  device.addEventListener('uncapturederror', (ev) => {
+  device.addEventListener('uncapturederror', (ev: GPUUncapturedErrorEvent) => {
```
greggman, line 59: You shouldn't need to add the type

```
npm install --save-dev @webgpu/types@latest
```

kainino0x: oh, right! I think I can easily update the types since we're not strictly enforcing them anyway

kainino0x: I moved the unrelated nits together with #512.

kainino0x: And rolled types in #511.

beaufortfrancois (approved):

PR #513, opened 2025-05-07
Prevent image flash when switching between texture formats

beaufortfrancois:
An image flash occurs when switching between texture formats. I believe it happens because writeTexture may not be complete causing the canvas to use incomplete texture data.

### DIFF
--- a/sample/volumeRenderingTexture3D/main.ts
+++ b/sample/volumeRenderingTexture3D/main.ts
@@ -130,10 +130,11 @@ const uniformBuffer = device.createBuffer({
 });
 
 let volumeTexture: GPUTexture | null = null;
+let isVolumeTextureReady = false;
 
 // Fetch the image and upload it into a GPUTexture.
 async function createVolumeTexture(format: GPUTextureFormat) {
-  volumeTexture = null;
+  isVolumeTextureReady = false;
 
   const { blockLength, bytesPerBlock, dataPath, feature } = brainImages[format];
   const width = 180;
@@ -172,6 +173,8 @@ async function createVolumeTexture(format: GPUTextureFormat) {
     { bytesPerRow: bytesPerRow, rowsPerImage: blocksHigh },
     [width, height, depth]
   );
+  await device.queue.onSubmittedWorkDone();
+  isVolumeTextureReady = true;
 }
 
 await createVolumeTexture(params.textureFormat);
@@ -260,7 +263,7 @@ function frame() {
 
   const commandEncoder = device.createCommandEncoder();
   const passEncoder = commandEncoder.beginRenderPass(renderPassDescriptor);
-  if (volumeTexture) {
+  if (isVolumeTextureReady) {
     bindGroupDescriptor.entries[2].resource = volumeTexture.createView();
     const uniformBindGroup = device.createBindGroup(bindGroupDescriptor);
     passEncoder.setPipeline(pipeline);
--- a/sample/volumeRenderingTexture3D/main.ts
+++ b/sample/volumeRenderingTexture3D/main.ts
@@ -130,11 +130,10 @@ const uniformBuffer = device.createBuffer({
 });
 
 let volumeTexture: GPUTexture | null = null;
-let isVolumeTextureReady = false;
 
 // Fetch the image and upload it into a GPUTexture.
 async function createVolumeTexture(format: GPUTextureFormat) {
-  isVolumeTextureReady = false;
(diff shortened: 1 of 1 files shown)


sample/volumeRenderingTexture3D/main.ts
```diff
@@ -172,6 +173,8 @@ async function createVolumeTexture(format: GPUTextureFormat) {
     { bytesPerRow: bytesPerRow, rowsPerImage: blocksHigh },
     [width, height, depth]
   );
+  await device.queue.onSubmittedWorkDone();
```
greggman, line 176: You shound't need to `await device.queue.onSubmittedWorkDone();` here.

greggman (commented):
To fix the flash,

In `createVolumeTexture`, move the call to `createTexture` to after the stream has been decompressed.

The reason it flashes is because

1. it creates a blank texture
2. it asynchronously waits for data to decompress - may take multiple frames during which the texture is transparent black
3. it then finally writes the data into the texture

You should never need to call `onScheduledWorkDone`. It's sole purpose is ATM is throttling.

beaufortfrancois:
> To fix the flash,
> 
> In `createVolumeTexture`, move the call to `createTexture` to after the stream has been decompressed.

I think it's better to fail early with non-supported texture formats rather than downloading the resource and then fail.
Moreover, the user sees an error right away in the canvas.

> The reason it flashes is because
> 
> 1. it creates a blank texture
> 2. it asynchronously waits for data to decompress - may take multiple frames during which the texture is transparent black
> 3. it then finally writes the data into the texture
> 
> You should never need to call `onScheduledWorkDone`. It's sole purpose is ATM is throttling.

I'm happy to remove `onScheduledWorkDone()`. I naively thought it was to make sure `writeTexture()` was effective.

greggman:
> > To fix the flash,
> > In `createVolumeTexture`, move the call to `createTexture` to after the stream has been decompressed.
> 
> I think it's better to fail early with non-supported texture formats rather than downloading the resource and then fail. Moreover, the user sees an error right away in the canvas.

The code already exists at like 147 if the format is not supported.

beaufortfrancois:
> > > To fix the flash,
> > > In `createVolumeTexture`, move the call to `createTexture` to after the stream has been decompressed.
> > 
> > 
> > I think it's better to fail early with non-supported texture formats rather than downloading the resource and then fail. Moreover, the user sees an error right away in the canvas.
> 
> The code already exits at line 147 if the format is not supported.

Even though `createTexture()` can still fail for other reasons I agree it should be enough in this context. Thanks for the review! I've updated the PR.

greggman:
> > > > To fix the flash,
> > > > In `createVolumeTexture`, move the call to `createTexture` to after the stream has been decompressed.
> > > 
> > > 
> > > I think it's better to fail early with non-supported texture formats rather than downloading the resource and then fail. Moreover, the user sees an error right away in the canvas.
> > 
> > 
> > The code already exits at line 147 if the format is not supported.
> 
> Even though `createTexture()` can still fail for other reasons I agree it should be enough in this context. Thanks for the review! I've updated the PR.

You could change the code to

    temp = createTexture
    fetch
    decompress
    writeData
    volumeTexture = temp

If you wanted some other error related to texture creation to show up first.

sample/volumeRenderingTexture3D/main.ts
```diff
@@ -150,13 +150,6 @@ async function createVolumeTexture(format: GPUTextureFormat) {
     status.textContent = '';
   }
 
-  volumeTexture = device.createTexture({
-    dimension: '3d',
-    size: [width, height, depth],
-    format: format,
-    usage: GPUTextureUsage.TEXTURE_BINDING | GPUTextureUsage.COPY_DST,
-  });
```
kainino0x, line 158: Didn't we do this early before waiting for the stream so that it wouldn't keep rendering with the old format after the UI was showing the new format?

greggman: It used to skip calling `pass.draw` etc but it still called `context.getCurrentTexture` and still called `encoder.beginRenderPass` so effectively the canvas was cleared until the new texture loaded. The "flash"

kainino0x: Ohh I see, we have already set it to null to stop rendering. I thought we were preventing the black flash, but that's desirable, what this change prevents is a white flash caused by 0s for astc being a white image.

PR #514, opened 2025-05-22
Use GPUBuffer in GPUBindingResource

beaufortfrancois:
Spec PR: gpuweb/gpuweb#5193

### DIFF
--- a/sample/a-buffer/main.ts
+++ b/sample/a-buffer/main.ts
@@ -448,24 +448,15 @@ const configure = () => {
     entries: [
       {
         binding: 0,
-        resource: {
-          buffer: uniformBuffer,
-          label: 'uniforms',
-        },
+        resource: uniformBuffer,
       },
       {
         binding: 1,
-        resource: {
-          buffer: headsBuffer,
-          label: 'headsBuffer',
-        },
+        resource: headsBuffer,
       },
       {
         binding: 2,
-        resource: {
-          buffer: linkedListBuffer,
-          label: 'linkedListBuffer',
-        },
+        resource: linkedListBuffer,
       },
       {
         binding: 3,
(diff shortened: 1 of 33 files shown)


beaufortfrancois:
FYI It requires https://github.com/gpuweb/types/pull/182

sample/a-buffer/main.ts
```diff
@@ -448,24 +448,15 @@ const configure = () => {
     entries: [
       {
         binding: 0,
-        resource: {
-          buffer: uniformBuffer,
-          label: 'uniforms',
```
kainino0x, line 453: oops that we had these labels in here at all, they don't do anything!

since we can't land this for a while, would you mind opening a separate PR to fix this?

beaufortfrancois: Sure. Here's https://github.com/webgpu/webgpu-samples/pull/515.

beaufortfrancois: I've fixed conflicts due to #515

kainino0x (approved):
LGTM but marking draft as we need to wait until all browsers support this.

mwyrzykowski:
> @jimblandy @mwyrzykowski Are there existing browser bugs/tickets I can follow to keep track of those changes?

Yes, WebKit bug is https://bugs.webkit.org/show_bug.cgi?id=294006 @beaufortfrancois

greggman:
@mwyrzykowski, my Mac had not told me to upgrade yet 😄 .  I updated to MacOS 26.2 and can verify that Safari 26.2 supports this.

tangential question: Safari TP no longer supports WebGPU on MacOS 15 right, or at least it stopped working for me. My work machine is still MacOS 15.7.3, my personal machine is now 26.2. I have not been able to use Safari TP on MacOS 15 to test WebGPU for a month or 2. Is that expected?

As for this PR. We might want to give it a couple of weeks to let more people update to 26.2 before breaking it for people still on < 26.2.

mwyrzykowski:
> tangential question: Safari TP no longer supports WebGPU on MacOS 15 right, or at least it stopped working for me. My work machine is still MacOS 15.7.3, my personal machine is now 26.2. I have not been able to use Safari TP on MacOS 15 to test WebGPU for a month or 2. Is that expected?

That's correct, it was disabled with https://bugs.webkit.org/show_bug.cgi?id=299237 since we removed some underlying framework workarounds for OS versions < 26.0

> As for this PR. We might want to give it a couple of weeks to let more people update to 26.2 before breaking it for people still on < 26.2.

Sounds good to me 👍

PR #520, opened 2025-06-23
Add WebGPUCompatibilityMode origin trial tokens

beaufortfrancois:
This PR adds tokens for the [WebGPU compatibility mode origin trial](https://developer.chrome.com/origintrials/#/view_trial/1489002626799370241):
- one for https://webgpu.github.io (production URL) expiring April 21, 2026 
- one for http://localhost:8080 (development URL) expiring April 21, 2026

image

### DIFF
--- a/sample/a-buffer/index.html
+++ b/sample/a-buffer/index.html
@@ -4,6 +4,16 @@
     <meta charset="utf-8">
     <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
     <title>webgpu-samples: a-buffer</title>
+    <!-- WebGPUCompatibilityMode origin token for https://webgpu.github.io expiring April 21, 2026 -->
+    <meta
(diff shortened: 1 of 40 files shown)


beaufortfrancois:
FYI @SenorBlanco

greggman (approved):

SenorBlanco:
This doesn't seem to be working. Compatibiilty mode still seems to require enabling "WebGPU Developer Features" in chrome://flags.

I'm not sure if this is the issue, but when I register for an origin trial, the tokens I get end in a single equal-sign (=). The sample site token added here ends in a double-equals (==), and the localhost token has none.

SenorBlanco:
(Note that "WebGPU Developer Features" is exposed in chrome://flags, while "WebGPU Experimental Features" is not. So the former is easy to enable in Canary while the latter requires command-line flags.)

PR #521, opened 2025-07-10
Use texture shorthands instead of view when possible

beaufortfrancois:
Spec PR: https://github.com/gpuweb/gpuweb/pull/5228

Note that support is available only in Chrome Canary[ for now](https://chromiumdash.appspot.com/commit/d172c015da875206fc746701314cbfce77efc3db).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -31,7 +31,7 @@
         "@types/showdown": "^2.0.6",
         "@types/stats.js": "^0.17.4",
         "@typescript-eslint/eslint-plugin": "^7.18.0",
-        "@webgpu/types": "^0.1.61",
+        "@webgpu/types": "^0.1.64",
         "chokidar": "^4.0.3",
         "eslint": "^8.57.1",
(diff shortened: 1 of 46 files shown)


greggman (approved):
lgtm - I assume you'll wait to merge until the other browsers support this?

greggman:
FYI, Safari does not support using a canvas texture as a color target if you don't call `createView()` as of Safari 26.2 so we can't merge this yet

Kangz:
@mwyrzykowski does Safari support using textures instead of views, including for the canvas texture now?

mwyrzykowski:
> @mwyrzykowski does Safari support using textures instead of views, including for the canvas texture now?

It does except resolve textures as @greggman noticed https://bugs.webkit.org/show_bug.cgi?id=313159

The resolve path is corrected too but only in Safari Technology Preview 243 and later

Kangz:
Thanks for the confirmation! Hopefully it makes it in the next stable Safari release!

PR #528, opened 2025-07-16
Revert "Don't use Compatibility Mode till Chrome bug is fixed (#523)"

SenorBlanco:
This reverts commit b28ddf74c34cd9b6d0db29d1974022554caca5f1.

Bug was [fixed in Dawn](https://dawn-review.googlesource.com/c/dawn/+/252698) and rolled into [Chrome 1487302](http://crrev.com/1487302), shipping in today's Canary.


beaufortfrancois (approved):

yashrajbharti:
Glad to see it's fixed! 🎉

greggman:
Doesn't the fix need to hit stable before this it's reverted?

PR #530, opened 2025-07-17
Feat/cursor control

### LINKED ISSUE #527: Sample Wanted: Cursor-Responsive Shader Interaction
I’d like to propose a cursor-interaction WebGPU sample that demonstrates fragment shader logic reacting to cursor input in a visual way.

## Core idea (Basic sample):
A single 3D model where the fragment shader changes texture/material under the mouse cursor. This would teach:
- Uniforms or buffers for passing cursor position
- Simple fragment logic for region-based effects
- UI-to-shader pipeline interaction

## Optional advanced extension
This may be too much for a sample but we can -
Stack two models and use the same shader logic to "reveal" one under the other, like a square-cut mask beneath the cursor.

### Example / Inspiration

https://github.com/user-attachments/assets/bb1cc42c-148d-4588-b2c9-90705439bd1d

(Credits: [Amin Ankward](https://www.linkedin.com/in/amin-ankward-djian-9886b517b/))

But we can keep this separate unless it’s of interest.

## Why?

There aren’t many visual WebGPU samples showing real-time interaction with shaders. This could make fragment shaders feel more intuitive for beginners, while giving room for deeper exploration such as this reveal effect advanced use case.

UPDATE: We now have two options to choose from #530 and #534 (happy to move forward with the one team aligns with! :smile:)

yashrajbharti:
## Description

This example shows interactive stencil masking in WebGPU using a single 2D plane that follows the cursor to create a dynamic cutout. A rotating torus is revealed only within this masked area, while a sphere remains visible outside it. The effect uses multi-pass rendering and stencil buffer logic to control visibility based on cursor interaction.

https://github.com/user-attachments/assets/cd7d59a2-0f54-4cbf-a89a-2320fd38dbe4

fixes #527

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+  "editor.formatOnSave": false
+}
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,3 +1,0 @@
-{
-  "editor.formatOnSave": false
-}
--- a/sample/cursorControl/index.html
+++ b/sample/cursorControl/index.html
@@ -1,0 +1,40 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta charset="utf-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
+    <title>webgpu-samples: cursor control</title>
+    <!-- WebGPUCompatibilityMode origin token for https://webgpu.github.io expiring April 21, 2026 -->
+    <meta
+      http-equiv="origin-trial"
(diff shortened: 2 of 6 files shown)


yashrajbharti:
Thanks for the suggestion. I’m open to moving in that direction! I’m slightly more inclined to keep this a basic, standalone example where the mouse interaction is more prominent and easier to follow.

The existing [stencilMask](https://webgpu.github.io/webgpu-samples/?sample=stencilMask) sample already does quite a bit with the six faces cube and various stencil values, making it more graphics-technique focused. In contrast, this cursor cutout sample leans toward demonstrating WebGPU’s interaction with DOM events, highlighting how a web platform cursor can influence 3D content in a minimal, less noisy setup.

That said, I’m happy to hear thoughts on pros/cons of both approaches and align with what feels most valuable for the examples!

PR #566, opened 2026-05-21
helloTriangleMSAA: avoid creating a texture every frame

### LINKED ISSUE #565: Hello Triangle MSAA creates a texture on every frame and doesn't destroy it, causing OOM in many configs
See https://matrix.to/#/!ZSOHTEPDbwuEgSJwYw:matrix.org/$H8TPgE0c6sxH0B8NO0ghr0C4F0D_iIN19uFtEr7Mb0Q?via=matrix.org&via=mozilla.org&via=krosse.krabbe.pizza

Unless i'm missing something the texture can be created only once? But otherwise it should be destroyed after submit.

kainino0x:
Fixes #565

### DIFF
--- a/sample/helloTriangleMSAA/main.ts
+++ b/sample/helloTriangleMSAA/main.ts
@@ -55,26 +55,42 @@ const pipeline = device.createRenderPipeline({
   },
 });
 
+const getMSAATempTextureView = (() => {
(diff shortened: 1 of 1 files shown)


beaufortfrancois (approved):
Thank you for fixing this!

Issue #350, opened 2024-01-24
Add small compute examples illustrating new WGSL primitives for AI


kenrussell:
@beaufortfrancois requested that some small examples be published here showing how to use the new WGSL primitives aimed at AI/ML workloads: `shader-f16`, DP4A, and soon, subgroups. Could we consider this?

Not sure what would be the most compelling - perhaps something with some visual output, and running a microbenchmark against the fallback WGSL code, assuming the feature is actually supported?

kenrussell:
CC @dneto0

dneto0:
For dp4a:
* They're good for ML image processing: segmentation, object identification.  But it's involved to put an ML model in the samples.
* Maybe we can do something really simple, like sobel filter on a grayscale image.

kainino0x:
dp4a accelerates matrix multiplication right? Even a basic matrix multiplication sample could be enough for a sample. The sample could even just display some text. But sobel or any other simple convolution would make it more compelling.

kainino0x:
Also it should have a toggle to enable/disable dp4a and hopefully see some performance improvement.

cmhhelgeson:
Is dp4a available in the current version of WebGPU?

austinEng:
dp4a is available starting in Chromium M123. So today, that would be Chrome Beta and newer.

cmhhelgeson:
If possible, I'd like to try my hand at this issue, at least for the next week (sorry about the timeline, day job is gonna day job). Sobel filter is a good place to start I think.

cmhhelgeson:
As for as I'm aware, WGSL only allows you to access 2d texture pixels as u32s or f32s. When a r8uint is read in as a 32 bit texture in WGSL, are the 8 bit values converted to 32 bit values or are four 8 bit values packed into a single 32 bit value?

cmhhelgeson:
Just want to make sure I understand the assignment, the intended use of dp4a here. Instead of writing, say, this for our sobel filter.... (below is in pseudo-wgsl)

```
(Pixels loaded from texture using textureLoad and global_invocation_id
let result =  1 * pixel1.r
        + 2 * pixel2.r
        + 1 * pixel 3. r
        - 1 * textureLoad(inputTexture, vec2<u32>(id.x + 1, id.y - 1), 0).r

textureStore(output, id.xy, result)
```

We should do something like this?

```
[...]

PR #569, opened 2026-06-23
Add Vision Transformer sample with attention visualization

lyonsno:
Runs DeiT-Tiny (5.7M params) inference entirely in WebGPU compute shaders to classify images, and visualizes attention maps as interactive heatmap overlays showing which image patches the model focuses on.

The sample is organized around the transformer compute stages: patch embedding, layer normalization, multi-head attention (Q/K/V projections, scaled dot-product scores, softmax, weighted sum), MLP with GELU, and residual connections. Each compute shader is self-contained with at most 7 bindings per bind group.

Model weights are int8 quantized (5.8MB committed binary). The quantized weights are dequantized to fp32 during loading. An offline Python converter (`tools/convert_deit_weights.py`) generates the weight file from the HuggingFace model; it is not needed to run the sample. If maintainers prefer the weights hosted externally instead of committed, I can move them.

Third-party attribution is in `sample/visionTransformer/THIRD_PARTY_NOTICES.md` (model: Meta DeiT Apache-2.0, images: Unsplash).

Related to #350. This demonstrates transformer building blocks in WebGPU compute but does not specifically exercise DP4A, shader-f16, or subgroups; those would make good follow-up primitive-focused samples.

I'm happy to make any changes needed. Please let me know if the scope or asset size is a concern.


kainino0x:
For us to merge any sample in this repository it needs to be maintainable by the repo maintainers. This is not a _ton_ of code but I do think it's a bit too large for us to maintain. It would also be hard for us to review for idiomatic usage, plus the licensing sounds a bit complex.

I would recommend hosting it yourself on GitHub Pages, and submitting that as an embedded version in the "External Samples" section which generally we're happy to add stuff to.

lyonsno:
Thanks, that makes sense. I agree this is a better fit for External Samples than for the in-repo sample set given the maintenance and review surface.

I’ve moved it into a standalone repo and deployed it on GitHub Pages:

- Demo: https://lyonsno.github.io/webgpu-vit-attention/
- Repo: https://github.com/lyonsno/webgpu-vit-attention

I’ll close this PR and open a smaller PR adding it to the External Samples section instead.

lyonsno:
Closing in favor of the smaller External Samples PR path discussed above.

PR #572, opened 2026-09-16
Add packed integer dot product sample

lyonsno:
Adds a small `dot4I8Packed` compute example toward #350.

The sample packs four signed 8-bit components into each `u32`, computes four example dot products, and reads the results back from the GPU. A dat.gui selector displays the input vectors and result for each example.

It requires `packed_4x8_integer_dot_product` and displays a requirement message when that WGSL language feature is unavailable. There is one packed compute pipeline, with no scalar fallback or in-sample result validation.

### Testing

- Lint, source typecheck, and the Node 20.20.2 build pass. The build emits existing typed-array warnings in unrelated samples.
- Chrome 153.0.8010.48, headless with `--enable-unsafe-webgpu`, on an Apple Metal adapter: all four examples returned the expected results, and the dat.gui selector was exercised.
- Desktop and narrow-screen layouts inspected. Injecting feature unavailability displayed the requirement message without requesting an adapter or compiling a shader.

No new dependencies or external assets.


test/packedIntegerDotProduct.test.mjs
kainino0x, line 1: this code should not be so complicated that it needs tests

sample/packedIntegerDotProduct/reference.ts
```diff
@@ -0,0 +1,117 @@
...
+  return {
+    requestedRoute,
+    effectiveRoute,
+    packedLanguageFeatureSupported,
+    fallbackReason:
+      requestedRoute === 'packed' && !packedLanguageFeatureSupported
+        ? `WGSL language feature '${packedDotLanguageFeature}' is unavailable`
+        : null,
+  };
+}
```
kainino0x: this seems excessively complicated, also I don't think we should have fallback in a sample, that makes it confusing about what implementation you're looking at

sample/packedIntegerDotProduct/reference.ts
```diff
@@ -0,0 +1,117 @@
...
+      `partial output: expected ${expected.length} results, received ${actual.length}`
+    );
+  }
+  for (let index = 0; index < expected.length; ++index) {
+    if (actual[index] !== expected[index]) {
+      throw new Error(
+        `result ${index} mismatch: expected ${expected[index]}, received ${actual[index]}`
+      );
+    }
+  }
```
kainino0x: this is a sample not a test, don't check results as it makes the total amount of code too much for the sample to be easily readable by a learner

sample/packedIntegerDotProduct/index.html
kainino0x, line 1: please use dat.gui like other samples instead of implementing custom UI

kainino0x (changes requested):

sample/packedIntegerDotProduct/main.ts
```diff
@@ -0,0 +1,105 @@
...
+  const gui = new GUI();
+  gui
+    .add(settings, 'example', {
+      'Mixed signs': 0,
+      'Minimum signed bytes': 1,
+      'Maximum signed bytes': 2,
+      'Positive components': 3,
+    })
+    .onChange(showResult);
+  showResult();
```
kainino0x: On second thought if we're just going to precompute all the results once, we should just show them all, which I've done.

sample/packedIntegerDotProduct/main.ts
```diff
@@ -0,0 +1,105 @@
...
+    [-128, -128, -128, -128],
+    [127, 127, 127, 127],
+    [1, 2, 3, 4],
+  ];
+  const rhs = [
+    [-5, 6, -7, 8],
+    [-128, -128, -128, -128],
+    [127, 127, 127, 127],
+    [4, 3, 2, 1],
+  ];
```
kainino0x: I've paired these so it's easier to add more cases.

kainino0x (approved):

kainino0x:
Decided to add back a GUI to let people actually play with the inputs of the function, and remove the batched execution because there's just one input. It's a really barebones sample so will be interested still in replacing it at some point.

